PR #22, opened 2022-11-23
adjust to new render templates config

h4ck3rk3y:
Depends on https://github.com/kurtosis-tech/kurtosis/pull/448


h4ck3rk3y:
This won't pass till we build & release a new version of CLI after the above PR

src/shared_utils/shared_utils.star
```diff
@@ -3,7 +3,7 @@ UDP_PROTOCOL = "UDP"
```
mieubrisse, line 3: Wait how did the docs check PR pass, given that there's no changelog update??

mieubrisse (approved):

src/shared_utils/shared_utils.star
h4ck3rk3y: i think docs-checker 0.2.0 was buggy, moved to 0.2.1 and it works (tested that it failed on no logs)

Issue #19, opened 2022-11-22
Perhaps rename this to eth2-package instead of eth2-module


h4ck3rk3y:

mieubrisse:
+1 to this, per following the new Python-like convention

h4ck3rk3y:
Closing this in favor of https://github.com/kurtosis-tech/kurtosis/issues/412

PR #26, opened 2022-11-24
convert module -> package

h4ck3rk3y:


h4ck3rk3y:
Closes #19

types.proto
```diff
@@ -16,15 +16,15 @@ message ModuleInput {
 	//  - only Ethereum network (EL and CL nodes) will be launched. Nothing else (no transaction spammer)
 	//  - params for the CL nodes will be ignored (e.g. CL node image, CL node extra params)
 	// This is a hack - it's not very elegant - but this is a commonly-requested feature
-	// The longterm solution is making the module trivial to decompose so we don't need flags like this; we're working
+	// The longterm solution is making the package trivial to decompose so we don't need flags like this; we're working
```
leoporoli, line 19: should we rename the message name also `ModuleInput` ? or is this the expect name in APIC?

leoporoli (approved):

types.proto
h4ck3rk3y: Its set to that in APIC, Guillaume will be removing proto completely.

leoporoli: ah! ok, thanks

PR #39, opened 2022-12-09
removed the args flag in favor of args arg

h4ck3rk3y:


README.md
```diff
@@ -180,10 +180,10 @@ Then, run the dev loop:
 1. Make your code changes
 1. Rebuild and re-run the package by running the following from the root of the repo:
    ```bash
-   kurtosis run . --args "{}"
```
gbouv, line 183: Isn't it equivalent to `kurtosis run .`?

gbouv (approved):

README.md
h4ck3rk3y: yes i think kevin wanted to highlight it can be changed when he wrote the preivous readme.

PR #47, opened 2022-12-15
added support for application protocol to eth2 package

Peeeekay:
Screen Shot 2022-12-15 at 12 24 13 PM


src/shared_utils/shared_utils.star
```diff
@@ -1,7 +1,7 @@
 TCP_PROTOCOL = "TCP"
 UDP_PROTOCOL = "UDP"
-
-
+HTTP_APPLICATION_PROTOCOL = "http"
+OPTIONAL_APPLICATION_PROTOCOL = ""
```
mieubrisse, line 4: I probably would have called this `NOT_PROVIDED_APPLICATON_PROTOCOL` or something like this, since the "optional" actually pertains to the applicationProtocol arg and not this particular constant

mieubrisse (approved):

PR #65, opened 2022-12-20
fix!: Use plan object

h4ck3rk3y:
BREAKING-CHANGE: Uses the `plan` object. Users will have to update their Kurtosis CLI to >= 0.63.0 and restart the engine

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,7 +40,14 @@ workflows:
       # -- PR check jobs ------------------------------------------
       - kurtosis-docs-checker/check-docs:
           should-check-changelog: false
-          markdown-link-check-config-json: "{}"
+          markdown-link-check-config-json: |
+            {
+                "ignorePatterns": [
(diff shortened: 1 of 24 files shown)


.circleci/config.yml
```diff
@@ -40,7 +40,14 @@ workflows:
...
           should-check-changelog: false
-          markdown-link-check-config-json: "{}"
+          markdown-link-check-config-json: |
+            {
+                "ignorePatterns": [
+                    {
+                        "pattern": "https://github.com/kurtosis-tech/eth2-package"
+                    }
+                ]
+            }
```
gbouv, line 50: why are we doing that?

gbouv (approved):

.circleci/config.yml
h4ck3rk3y: the release please thing links to itself, to compare between the last released versions and the docs-checker complains about it

PR #83, opened 2023-03-30
refactor: Move args from struct to dict

victorcolombo:


h4ck3rk3y (approved):
can we test this one with some arg passing manually. this one is so complex

PR #89, opened 2023-04-25
refactor: added custom port wait config for lighthouse validator

leoporoli:


leoporoli:
@h4ck3rk3y The build and CircleCi are failing now because the automatic port wait feature does not release yet, but it will work when we release this feature, I've already tested it on my local.

h4ck3rk3y (approved):
LGTM! Merge after the release :)

PR #92, opened 2023-05-04
fix: geth failing after ethash package removal

parithosh:
chore: The geth client requires a new field in its genesis.json file, this field specifies that the merge has already happened and would allow the client to start post-merge. The update of the `ethereum-genesis-generator` addresses this.


leoporoli:
closing it because it was merged here: https://github.com/kurtosis-tech/eth2-package/pull/93

PR #99, opened 2023-07-03
fix: Switch default images to latest

parithosh:
This should prevent config drift, some images we used as defaults are months old at this point.

### DIFF
--- a/src/package_io/parse_input.star
+++ b/src/package_io/parse_input.star
@@ -1,16 +1,16 @@
 DEFAULT_EL_IMAGES = {
-	"geth": "ethereum/client-go:v1.11.5",
+	"geth": "ethereum/client-go:latest",
 	"erigon": "thorax/erigon:devel",
-	"nethermind": "nethermind/nethermind:1.14.0",
+	"nethermind": "nethermind/nethermind:latest",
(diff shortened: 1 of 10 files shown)


src/package_io/parse_input.star
```diff
@@ -1,16 +1,16 @@
...
 	"besu": "hyperledger/besu:develop"
 }
 
 DEFAULT_CL_IMAGES = {
-	"lighthouse": "sigp/lighthouse:v3.5.0",
-	"teku":       "consensys/teku:23.1",
+	"lighthouse": "sigp/lighthouse:latest",
+	"teku":       "consensys/teku:latest",
 	"nimbus":     "statusim/nimbus-eth2:multiarch-latest",
 	"prysm":    "prysmaticlabs/prysm/beacon-chain:latest,prysmaticlabs/prysm/validator:latest",
```
h4ck3rk3y, line 12: lets change these to `prysmaticlabs/prysm-beacon-chain:latest,prysmaticlabs/prysm-validator:latest`

h4ck3rk3y (approved):

src/package_io/parse_input.star
parithosh: just pushed the change

PR #100, opened 2023-07-07
fix: Nimbus can't run when slot time is below 12s

parithosh:


src/package_io/parse_input.star
```diff
@@ -51,7 +52,8 @@ def parse_input(input_args):
 
 		if index == 0 and el_client_type in (BESU_NODE_NAME, NETHERMIND_NODE_NAME):
 			fail("besu/nethermind cant be the first participant")
-		
+		if cl_client_type in (NIMBUS_NODE_NAME) and (result["network_params"]["seconds_per_slot"] < 12):
```
h4ck3rk3y, line 55: should we add this to `eth-network-package` as weell?

h4ck3rk3y: perhaps longer term we should remove the parsing here. maybe @tedim52 can get to it when he comes back in a few weeks 😅

h4ck3rk3y (approved):

src/package_io/parse_input.star
parithosh: sounds like a good plan. should i go ahead and merge this and we create an issue to move the logic downstream?

h4ck3rk3y: Happy for you to merge this. I'll add this line to eth-network-package anyway.

Create an issue for us to merge the logic together so that we don't have two different parsers in either of the repositories!

h4ck3rk3y: https://github.com/kurtosis-tech/eth-network-package/pull/44

PR #101, opened 2023-07-07
fix: Update starting genesis state to start with capella

barnabasbusa:


barnabasbusa:
Close in favor of #107

PR #108, opened 2023-07-14
feat: make it possible to have capella on epoch 0 or non 0

barnabasbusa:

feat: set capella to epoch 0


barnabasbusa:
Once https://github.com/kurtosis-tech/eth-network-package/pull/61 merged this can also be merged

src/package_io/parse_input.star
```diff
@@ -101,9 +101,6 @@ def parse_input(input_args):
 	if result["network_params"]["genesis_delay"] == 0:
 		fail("genesis_delay is 0 needs to be > 0 ")
 
-	if result["network_params"]["capella_fork_epoch"] == 0:
```
h4ck3rk3y, line 104: can we add an if condition that checks if mev is enabled and complain if capella epoch is less than 1

h4ck3rk3y (approved):

PR #109, opened 2023-07-17
feat: add pylint

barnabasbusa:


barnabasbusa:
reverted to use spaces instead of tabs due to: 
https://github.com/psf/black/issues/47

PR #110, opened 2023-07-19
fix: use eth maintained tx-fuzz

h4ck3rk3y:


src/transaction_spammer/transaction_spammer.star
```diff
@@ -10,19 +10,12 @@ def get_config(prefunded_addresses, el_client_context):
...
-	for prefunded_address in prefunded_addresses:
-		private_keys_strs.append(prefunded_address.private_key)
-		address_strs.append(prefunded_address.address)
-
-	comma_separated_private_keys = ",".join(private_keys_strs)
-	comma_separated_addresses = ",".join(address_strs)
 	return ServiceConfig(
 		image = IMAGE_NAME,
 		cmd = [
-			"http://{0}:{1}".format(el_client_context.ip_addr, el_client_context.rpc_port_num),
```
h4ck3rk3y, line 22: is there a way to give the offical txfuzz multiple private keys?

h4ck3rk3y: https://github.com/kurtosis-tech/tx-fuzz/blob/main/cmd/livefuzzer/main.go#L261-L266

allowed specifiying keys and addresses

parithosh: do we need to support multiple keys? tx-fuzz does this automatically, it basically "creates" 10 addresses, funds them and then performs transactions with those.

h4ck3rk3y: oh sure, no need then I think 👍

parithosh (approved):

PR #112, opened 2023-07-31
feat: counting by summing each participant

adschwartz:
This PR changes the way we count the actual number of validators by summing `count`s for each participant and multiplying by the number of validator keys per node.

Resolves https://github.com/kurtosis-tech/eth2-package/issues/111

feat: counting by summing each participant

### DIFF
--- a/src/package_io/parse_input.star
+++ b/src/package_io/parse_input.star
@@ -45,7 +45,7 @@ def parse_input(input_args):
 				participants.append(new_participant)
 			result["participants"] = participants
 
-	total_participant_count = 1
+	total_participant_count = 0
 	# validation of the above defaults
 	for index, participant in enumerate(result["participants"]):
 		el_client_type = participant["el_client_type"]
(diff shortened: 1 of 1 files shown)


src/package_io/parse_input.star
```diff
@@ -45,6 +45,7 @@ def parse_input(input_args):
 				participants.append(new_participant)
 			result["participants"] = participants
 
+	total_participant_count = 1
```
mieubrisse, line 48: @adschwartz shouldn't this start at 0 rather than 1?

adschwartz: Yeah the default needs to be `1` but only if there no `count`s defined. The way I did it was wrong so fixing that. Thank you!

mieubrisse (approved):

barnabasbusa (approved):
LGTM

src/package_io/parse_input.star
```diff
@@ -100,10 +103,13 @@ def parse_input(input_args):
 	if result["network_params"]["deneb_fork_epoch"] == 0:
 		fail("deneb_fork_epoch is 0 needs to be > 0 ")
 
-	required_num_validtors = 2 * result["network_params"]["slots_per_epoch"]
-	actual_num_validators = len(result["participants"]) * result["network_params"]["num_validator_keys_per_node"]
-	if required_num_validtors > actual_num_validators:
-		fail("required_num_validtors - {0} is greater than actual_num_validators - {1}".format(required_num_validtors, actual_num_validators))
+	if total_participant_count < 1:
```
barnabasbusa, line 106: I don't particularly see a scenario where total_participant_count could be less than 1. If there are no participants, there wouldn't be any test running at all.

PR #124, opened 2023-08-14
docs: added note about parallel keystore generation

h4ck3rk3y:


README.md
```diff
@@ -150,6 +150,9 @@ To configure the package behaviour, you can modify your `eth2-package-params.yam
          //  The deneb for epoch -- arbitrarily large while we sort out https://github.com/kurtosis-tech/eth-network-package/issues/42 this will take 53~ hours for now
          "deneb_for_epoch": 500,
 
+         // Parallelizes keystore generation so that each node has keystores being generated in their own container
+         // Use against large clusters only
```
leeederek, line 154:
```suggestion
         // This will result in a large number of containers being spun up than normal. We advise users to only enable this on a sufficiently large machine or in the cloud as it can be resource consuming on a single machine.
```

leeederek (approved):
looks good, just expanded on the comment you made for the `parallel_keystore_generation` field

PR #125, opened 2023-08-14
feat: add light-beaconchain-explorer

barnabasbusa:

feat: add light-beaconchain-explorer


light-beaconchain-config.yaml
```diff
@@ -0,0 +1,78 @@
...
+  enabled: true # Enable or disable to web frontend
+  debug: false
+  minimize: false # minimize html templates
+
+  # Name of the site, displayed in the title tag
+  siteName: "Beaconchain Light"
+  siteSubtitle: "Ephemery"
+
+  # link to EL Explorer
+  ethExplorerLink: "https://explorer.ephemery.dev/"
```
parithosh, line 31: change to something else pls

barnabasbusa:
Indexing seems to be not working right now. 
You can run this branch with `kurtosis run github.com/kurtosis-tech/eth2-package@bbusa/light-beaconchain`

pk910:
Thanks for that command :)
Spinning up a local devnet that way is impressively easy.

I've fixed various issues in the explorer that in combination lead to that bad result.
The main issue has probably been, that the indexer thread silently died when the first RPC call wasn't successful.

I've published a new bugfix release for the explorer, so a working image is now available via the `pk910/light-beaconchain-explorer:latest` image too.

PR #137, opened 2023-08-18
ci: add more test cases and rename tests

barnabasbusa:


.circleci/config.yml
```diff
@@ -48,137 +38,25 @@ jobs:
...
-                - main
-      - run_starlark:
-          context:
-            # This pulls in KurtosisBot's Github token, so that we can read from our private repos
-            - github-user
-          filters:
-            branches:
-              ignore:
-                - main
+      - run_starlark
```
h4ck3rk3y, line 47: adding this comment so i don't forget but this test should be nightly so behind the 

`when: << pipeline.parameters.should-enable-check-latest-version-workflow >>` flag 

don't worry about it now. this is anote to self

PR #145, opened 2023-08-24
feat: skip validation for 0 count participants

joshieDo:
Useful for when wanting to test certain parameters and not wanting to completely remove incompatible clients from the configuration file.  Example: setting `seconds_per_slot: 2` forces us to remove every nimbus pair

### DIFF
--- a/src/package_io/parse_input.star
+++ b/src/package_io/parse_input.star
@@ -55,6 +55,9 @@ def parse_input(input_args):
 	total_participant_count = 0
 	# validation of the above defaults
 	for index, participant in enumerate(result["participants"]):
+		if participant["count"] == 0:
+		    continue
+
 		el_client_type = participant["el_client_type"]
 		cl_client_type = participant["cl_client_type"]
 


leeederek:
Hey @joshieDo thank you for opening a PR with your suggestion! Can you elaborate on your workflow? For example, are you setting the `seconds_per_slot: 2` after you've already spun up the network & wish to observe how the network behaves following that change?

h4ck3rk3y:
@joshieDo adding on to Derek's question - do you want to be able to spin up the nimbus pair or not? Or do you want the ability to have some configuration and set counts to 0, so that you don't have to remove it fully

joshieDo:
> Or do you want the ability to have some configuration and set counts to 0, so that you don't have to remove it fully

Yeah, exactly this.

>Can you elaborate on your workflow? For example, are you setting the seconds_per_slot: 2 after you've already spun up the network & wish to observe how the network behaves following that change?

No, no, just from scratch

leeederek:
@joshieDo can you elaborate on your workflow? I don't quite follow your sequence of operations. 

My understanding is that you're spinning up a network with Nimbus with `seconds_per_slot: 12` and then you want to change the `seconds_per_slot: 2` and spin up the same network again without Nimbus. 

If the above is true, then you can accomplish what you're describing without any changes to the eth2-package. You should be able to do this workflow by editing your `network_params.json` file to remove Nimbus and change `seconds_per_slot:` to `2` and re-run the same package.

joshieDo:
> If the above is true, then you can accomplish what you're describing without any changes to the eth2-package. You should be able to do this workflow by editing your network_params.json file to remove Nimbus and change seconds_per_slot: to 2 and re-run the same package.

It is true, it's more of a UX thing, where I don't want to be removing all the nimbus configuration (and adding it back again). nbd, feel free to close the PR! :sweat_smile:

leeederek:
Gotcha, no worries. We're totally cool with this change - just wanted to understand the rationale behind it a bit better. However, you'll likely need to make a similar change here too as we validate the nimbus arguments here as well: https://github.com/kurtosis-tech/eth-network-package/blob/main/package_io/input_parser.star#L50-L53 

We can approve if you make the change to the eth-network-package as well to enable your workflow, since the eth2-package depends on the vanilla eth network spun up by the eth-network-package.

h4ck3rk3y:
Closing this as there has been no action on this PR for 3 weeks. @joshieDo feel free to re-open if you are still keen on this!

PR #153, opened 2023-08-28
fix: use v2 endoint to get the head block

bharath-123:


parithosh (approved):

parithosh:
Approved and tested it locally, merging this PR now.

parithosh:
@bharath-123 you might need to rebase with main before this can be merged.

PR #159, opened 2023-08-29
fix: nightly tests that rely on etherejums get the right image

h4ck3rk3y:


barnabasbusa:
I'd prefer if we don't have two parsers. Lets import the parser from eth-network-client. Then this kind of issue would never happen.

barnabasbusa (approved):

PR #176, opened 2023-09-01
feat: add support for electra

barnabasbusa:


README.md
```diff
@@ -166,10 +191,13 @@ To configure the package behaviour, you can modify your `network_params.json` fi
...
         "preregistered_validator_keys_mnemonic": "giant issue aisle success illegal bike spike question tent bar rely arctic volcano long crawl hungry vocal artwork sniff fantasy very lucky have athlete",
+        // How long you want the network to wait before starting up
+        "genesis_delay": 120,
 
         // The epoch at which the capella and deneb forks are set to occur.
-        "capella_fork_epoch": 2,
-        "deneb_fork_epoch": 4
+        "capella_fork_epoch": 0,
+        "deneb_fork_epoch": 4,
+        "electra_fork_epoch": "None",
```
h4ck3rk3y, line 200: should this be `"None"` or should this be `null` (json)?

h4ck3rk3y (approved):

README.md
barnabasbusa: null json, you are right, fixing it now 

got the None from your mev one `    "mev_type": "None",` Should I use null ?

barnabasbusa: seems like both is fine.

h4ck3rk3y: The `None` is probably a mistake but if either work I am happy going with `null`

PR #179, opened 2023-09-04
fix: Kevin/postgres package upgrade

mieubrisse:
This uses the latest postgres-package, which has the unbundled args format for clarity


h4ck3rk3y (approved):
nice

PR #186, opened 2023-09-05
feat: added a reliable flooder

h4ck3rk3y:


args.json
```diff
@@ -0,0 +1,10 @@
+{
```
h4ck3rk3y, line 1: remove this before merge!!!!

src/mev_custom_flood/sender.py
```diff
@@ -0,0 +1,52 @@
...
+        time.sleep(3)
+        
+        w3.middleware_onion.add(construct_sign_and_send_raw_middleware(sender_account))
+
+        tx_hash = w3.eth.send_transaction({
+            "from": sender_account.address,
+            "value": 0x9184,
+            "to": receiver,
+            "data": "0xabcd",
+            "gasPrice": "0x9184e7",
```
parithosh, line 42: Any reason to not use the estimate gas call? that would ensure your tx uses the appropriate gas and leads to more predictable transactions

src/mev_custom_flood/sender.py
```diff
@@ -0,0 +1,52 @@
...
+logging.basicConfig(filename="/tmp/sender.log",
+                    filemode='a',
+                    format='%(asctime)s,%(msecs)d %(name)s %(levelname)s %(message)s',
+                    datefmt='%H:%M:%S',
+                    level=logging.DEBUG)
+
+
+def flood():
+    # this is the last prefunded address
+    sender = os.getenv("SENDER_PRIVATE_KEY", "17fdf89989597e8bcac6cdfcc001b6241c64cece2c358ffc818b72ca70f5e1ce")
```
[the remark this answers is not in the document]
h4ck3rk3y: @parithosh I take your point regarding private key usage but I don't want to leave a comment on eth-network-package about eth2-package, as eth-network-package shouldn't ideally have knowledge about its consumer

what I can do is in the README of eth2-package add a table showing current keys used

PR #188, opened 2023-09-08
feat: pass slots per epoch to mev-boost-relay

h4ck3rk3y:


src/mev_relay/mev_relay_launcher.star
```diff
@@ -17,7 +17,7 @@ NETWORK_ID_TO_NAME = {
 	"3":        "ropsten",
 }
 
-def launch_mev_relay(plan, mev_params, network_id, beacon_uris, validator_root, builder_uri, seconds_per_slot):
+def launch_mev_relay(plan, mev_params, network_id, beacon_uris, validator_root, builder_uri, seconds_per_slot, slots_per_epoch = 32):
```
barnabasbusa, line 20: any reason you set this to 32? 
32 is the default.

h4ck3rk3y: i set this to 32 here so that if there are any consumers of the package they don't get a breaking change and this continues to work for them

barnabasbusa (approved):

PR #192, opened 2023-09-12
refactor: call down to eth-network input parser

### LINKED ISSUE #161: dedupe parsing between eth2-package and eth-network-package
perhaps eth2 can just call down to eth-network?

leoporoli:
Fix #161


src/package_io/parse_input.star
```diff
@@ -164,61 +76,6 @@ def parse_input(plan, input_args):
 		mev_type=result["mev_type"],
 	), result
 
-def get_client_log_level_or_default(participant_log_level, global_log_level, client_log_levels):
```
leoporoli, line 167: this method is not used inside the package

PR #205, opened 2023-09-18
fix: parse input

leoporoli:


leeederek (approved):

src/package_io/parse_input.star
```diff
@@ -17,8 +19,11 @@ def parse_input(plan, input_args):
 
 	for attr in input_args:
 		value = input_args[attr]
+		# if its insterted we use the value inserted
```
leeederek, line 22: Typo here in the comment

leoporoli: fixed

PR #206, opened 2023-09-19
ci: support arm64

h4ck3rk3y:
this image is built on both amd64 and arm64, unlike the original. Will replace with something built & maintained by EthPandaOps @parithosh @barnabasbusa 

This should unblock Guillaume from EF


h4ck3rk3y:
^ will look at this tomorrow

leeederek (approved):
Good to me, so long as checks pass

PR #209, opened 2023-09-20
feat: add generic prometheus endpoints

cbermudez97:
- feat: add execution and additional metrics jobs to prometheus service
- feat: add beacon metrics gazer to prometheus
- fix: ignore not defined metrics info


src/prometheus/prometheus_launcher.star
```diff
@@ -64,7 +65,74 @@ def get_config(config_files_artifact_name):
...
+	labels,
+	scrape_interval = "",
+):
 	return {
-		"CLNodesMetricsInfo": cl_nodes_metrics_info
+		"Name": job_name,
+		"Endpoint": endpoint,
+		"MetricsPath": metrics_path,
+		"Labels": labels,
+		"ScrapeInterval": scrape_interval,
```
h4ck3rk3y, line 137: does setting ScrapeInterval to "" work? Or should we set it to a reasonable default?

h4ck3rk3y: Say 15s like below

static_files/prometheus-config/prometheus.yml.tmpl
```diff
@@ -1,16 +1,15 @@
...
-       - targets: ['{{ $clNode.url }}']
-   {{ end }}
-   - job_name: 'beacon-metrics-gazer'
-     metrics_path: '/metrics'
-     static_configs:
-      - targets: ['beacon-metrics-gazer:8080']
+  {{- range $job := .MetricsJobs }}
+  - job_name: "{{ $job.Name }}"
+    metrics_path: "{{ $job.MetricsPath }}"
+    {{- if $job.ScrapeInterval }}
```
h4ck3rk3y, line 7: oh you test if its non ""

h4ck3rk3y (approved):
LGTM thanks for contributing

1. Do we have Grafana dashboards for the EL metrics?
2. Have you tested if Grafana works properly for CL clients

src/prometheus/prometheus_launcher.star
cbermudez97: Its ignored inside the template if empty. In those cases uses default 15s. I guess adding 15s on default is fine too.

main.star
```diff
@@ -287,25 +294,28 @@ def run(plan, args={}):
             )
             plan.print("Succesfully light-beaconchain-explorer")
         elif additional_service == "prometheus_grafana":
```
h4ck3rk3y, line 296: ah i see there's an ordering issue - can we do the following; remove this if condition completely from inside the loop

and do something like this at the end; perhaps we could also add an `or

```
if "prometheus_grafana" in args_with_defaults.additional_services or "beacon_metrics_gazer" in args_with_defaults.additional_services:
     launch_it_like_we_do
```

h4ck3rk3y (approved):

main.star
```diff
@@ -83,6 +83,7 @@ def run(plan, args={}):
     prometheus_config_template = read_file(
         static_files.PROMETHEUS_CONFIG_TEMPLATE_FILEPATH
     )
+    prometheus_additional_metrics_jobs = []
```
h4ck3rk3y, line 86: temporarily adding a blocking comment while i am cleaning up the repo

main.star
cbermudez97: Okay, as you suggested, I will remove the condition and add the check only at the end. I think there is no need to add the `or` part since Prometheus can run independently of the `beacon_metrics_gazer`.

main.star
```diff
@@ -283,12 +285,17 @@ def run(plan, args={}):
...
+            beacon_metrics_gazer_prometheus_metrics_job = (
+                beacon_metrics_gazer.launch_beacon_metrics_gazer(
+                    plan,
+                    beacon_metrics_gazer_config_template,
+                    all_cl_client_contexts,
+                    args_with_right_defaults.participants,
+                    network_params,
+                )
+            )
+            prometheus_additional_metrics_jobs.append(
```
h4ck3rk3y, line 297: should we set `launch_prometheus_grafana` to true here as well? is this useful otherwise

h4ck3rk3y (approved):

PR #219, opened 2023-09-21
fix: rename light-beaconchain-explorer to dora-the-explorer

pk910:
The lightweight beaconchain explorer was renamed to `Dora the Explorer` (or just `Dora`).
This PR renames all explorer related variables & references accordingly.

PR #243, opened 2023-09-28
fix: rename light-beaconchain-explorer to dora-the-explorer & change db location

pk910:
The lightweight beaconchain explorer was renamed to `dora` (Dora the Explorer).
This PR renames all explorer related variables & references accordingly.

I've also changed the explorer db path to a local file instead of the `:memory:` placeholder.
The in-memory database caused issues as the content is being wiped when all db connections are being closed.
I've just used `/dora-database.sqlite` as path. Maybe some of you can give feedback if that's the right location for it :)

Replacement for #219, as the renaming has been reworked in a new branch


static_files/dora-config/config.yaml.tmpl
```diff
@@ -41,31 +37,27 @@ beaconapi:
       archive: true
 {{- end }}
   # local cache for page models
-  localCacheSize: 100 # 100MB
+  localCacheSize: 10 # 10MB
```
h4ck3rk3y, line 40: not sure what this is used for but is reducing this alright?

h4ck3rk3y (approved):

static_files/dora-config/config.yaml.tmpl
pk910: yea, the cache is used to preserve page models.
I've set it to a high limit to handle public testnet scenarios (so with probably hundreds of uses using the explorer at the same time)

but for a local only testnet 100MB is quite high, so I've reduced it to 10MB which is still enough to cache hundreds of page models :)
Lower cache does not really affect the explorer functionality. Page processing is quite fast, even with the localCache disabled at all (debug mode)

PR #257, opened 2023-10-02
refactor: using local relative locators instead of absolute

leoporoli:


h4ck3rk3y (approved):
trusting that these are all the replacements!

PR #261, opened 2023-10-04
fix: set MEV image to 0.26.0 and complain if capella is zero with MEV set to full

h4ck3rk3y:


.circleci/tests/mev.json
```diff
@@ -4,6 +4,7 @@
         "launch_custom_flood": true
     },
     "network_params": {
-        "seconds_per_slot": 3
+        "seconds_per_slot": 3,
```
h4ck3rk3y, line 7: THIS IS A BLOCKER FROM MERGE

This PR is just a discussion

barnabasbusa (approved):

PR #267, opened 2023-10-05
feat: Mev Deneb

avalonche:
using deneb images for testing kurtosis flow. 

Seeing blocks coming through proposed through the relay ~epoch 5. This does not support custom slot time.

### DIFF
--- a/network_params.json
+++ b/network_params.json
@@ -2,11 +2,11 @@
 	"participants": [
 		{
 			"el_client_type": "geth",
-			"el_client_image": "ethereum/client-go:latest",
+			"el_client_image": "ethpandaops/geth:lightclient-devnet-9-ae44519",
 			"el_client_log_level": "",
 			"el_extra_params": [],
 			"cl_client_type": "lighthouse",
-			"cl_client_image": "sigp/lighthouse:latest",
+			"cl_client_image": "ethpandaops/lighthouse:deneb-free-blobs-57edc0f",
 			"cl_client_log_level": "",
 			"beacon_extra_params": [],
 			"validator_extra_params": [],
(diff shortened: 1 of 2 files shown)


network_params.json
```diff
@@ -2,11 +2,11 @@
 	"participants": [
 		{
 			"el_client_type": "geth",
-			"el_client_image": "ethereum/client-go:latest",
+			"el_client_image": "ethpandaops/geth:lightclient-devnet-9-ae44519",
```
barnabasbusa, line 5: There is no need to override the defaults.

src/package_io/parse_input.star
```diff
@@ -1,5 +1,5 @@
 DEFAULT_EL_IMAGES = {
-    "geth": "ethereum/client-go:latest",
+    "geth": "ethpandaops/geth:lightclient-devnet-9-ae44519",
```
barnabasbusa, line 2: Same as the comment above.

src/package_io/parse_input.star
```diff
@@ -447,7 +448,7 @@ def enrich_mev_extra_params(parsed_arguments_dict, mev_prefix, mev_port, mev_typ
                 "el_client_image": parsed_arguments_dict["mev_params"][
                     "mev_builder_image"
                 ],
-                "cl_client_image": "sigp/lighthouse",
+                "cl_client_image": "ethpandaops/lighthouse:deneb-free-blobs-57edc0f",
```
barnabasbusa, line 451: We have now introduced a flag for `mev_builder_cl_image` so you can also override this.

barnabasbusa:
All the mev related images can be overritten in the local config file.

PR #268, opened 2023-10-06
feat: added another blob spamming tool (`goomy_blob`)

pk910:
Added Goomy the Blob Tool to the package.
With default parameters it will generate a low, but continuous amount of blobs (2-4 blobs per block).


main.star
```diff
@@ -234,6 +235,18 @@ def run(plan, args={}):
                 network_params.genesis_delay,
             )
             plan.print("Succesfully launched blob spammer")
+        elif additional_service == "goomy_blob":
```
h4ck3rk3y, line 238: we could also do this the way we do dora / full explorer so via `explorer_version`; we can add a `blob_explorer_type` perhaps

README.md
```diff
@@ -210,6 +210,12 @@ To configure the package behaviour, you can modify your `network_params.json` fi
         "tx_spammer_extra_args": []
     },
```
h4ck3rk3y, line 212: somewhere up / down there's a `launch_additional_services` arg; can we say how one can add `gloomy_blob` there

h4ck3rk3y (approved):

main.star
pk910: as discussed on discord, we'd like to proceed with the additional_services array

PR #275, opened 2023-10-08
feat: Adding 4788 deployment

parithosh:


src/mev_custom_flood/mev_custom_flood_launcher.star
```diff
@@ -1,12 +1,12 @@
 PYTHON_IMAGE = "python:3.11-alpine"
-CUSTOM_FLOOD_SREVICE_NAME = "mev-custom-flood"
+CUSTOM_FLOOD_SERVICE_NAME = "mev-custom-flood"
```
h4ck3rk3y, line 2: thanks!

src/eip4788_deployment/eip4788_deployment_launcher.star
```diff
@@ -0,0 +1,32 @@
+PYTHON_IMAGE = "python:3.11-alpine"
+EIP4788_DEPLOYMENT_SERVICE_NAME = "eip4788-contract-deployment"
+
+
+def deploy_eip4788_contract_in_background(plan, sender_key, receiver_key, el_uri):
```
h4ck3rk3y, line 5: I wonder if this can be extracted to something common between this and `sender.py`;

h4ck3rk3y: second thought, this script exit while the other script keeps running; maybe we can make the cmd

```
cmd=["/bin/sh", "-c", "pip install web3 && /tmp/sender.py"],
```

this will STOP the container when the script exits; the previous one needs the fancy no hup and such as that needs to keep running infinitely

main.star
```diff
@@ -95,6 +97,18 @@ def run(plan, args={}):
...
 
+    plan.print("Launching 4788 contract deployer")
+    if network_params.deneb_fork_epoch != 0:
+        el_uri = "http://{0}:{1}".format(
+            all_el_client_contexts[0].ip_addr, all_el_client_contexts[0].rpc_port_num
+        )
+        eip4788_deployment_module.deploy_eip4788_contract_in_background(
+            plan,
+            genesis_constants.PRE_FUNDED_ACCOUNTS[5].private_key,
+            "0x0B799C86a49DEeb90402691F1041aa3AF2d3C875",
```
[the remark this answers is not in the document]
parithosh: Its a presigned deterministic deployment address, you can never change it for the 4788 contract :)

src/eip4788_deployment/eip4788_deployment_launcher.star
parithosh: fixed, have a look

main.star
```diff
@@ -95,6 +97,18 @@ def run(plan, args={}):
         all_el_client_contexts.append(participant.el_client_context)
         all_cl_client_contexts.append(participant.cl_client_context)
 
+    plan.print("Launching 4788 contract deployer")
```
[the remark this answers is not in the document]
parithosh: fixed

main.star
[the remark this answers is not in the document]
parithosh: and the deployment is hardcoded in clients as well, so changing it will be a whole drama anyway

src/eip4788_deployment/eip4788_deployment_launcher.star
parithosh: tried it out, this approach blocks the exec phase until the contract deployment is done - which cannot happen until the chain has started. I'd prefer spinning it up and letting kurtosis' plan/exec phase finish and the deployment happens later on.

main.star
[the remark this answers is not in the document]
h4ck3rk3y: Oh i mean, take the "0x0B799C86a49DEeb90402691F1041aa3AF2d3C875" out and in a variable called "DEPLOYMENT_ADDRESS" on top and use that here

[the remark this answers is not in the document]
h4ck3rk3y: or if its not changeable at all, maybe it should be hardcoded in the `sender.py`

src/eip4788_deployment/eip4788_deployment_launcher.star
h4ck3rk3y: oh alright, makes sense to me.

h4ck3rk3y, line 5: perhaps this can be pulled out in a `run_python_script_in_background` method that gets shared between this and `mev_custom_flood`; don't have to do it right now but maybe worth trackign a TODO

parithosh: just more co-reuse. i'd toss a TODO for now and leave it at that

main.star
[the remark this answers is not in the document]
parithosh: fixed by removing receiver address

src/eip4788_deployment/eip4788_deployment_launcher.star
parithosh: issue here: https://github.com/kurtosis-tech/ethereum-package/issues/287

PR #278, opened 2023-10-09
fix: fix the tx_fuzzer params

kamilchodola:


src/package_io/parse_input.star
```diff
@@ -95,7 +89,9 @@ def parse_input(plan, input_args):
             )
         )
 
-    result["tx_spammer_params"] = get_default_tx_spammer_params()
+    if "tx_spammer_params" not in result:
+        result["tx_spammer_params"] = get_default_tx_spammer_params()
```
piwonskp, line 93: It'd overwrite all params, even those unspecified. If user does not specify all params, the command will fail. The code will prevent that: 

```suggestion
    result["tx_spammer_params"] = dict(get_default_tx_spammer_params(), **result["tx_spammer_params"])
```

kamilchodola: Applied, thanks!

src/package_io/parse_input.star
```diff
@@ -92,7 +91,10 @@ def parse_input(plan, input_args):
             )
         )
 
-    result["tx_spammer_params"] = get_default_tx_spammer_params()
+    result["tx_spammer_params"] = dict(
```
h4ck3rk3y, line 94: Nice find; to keep it consistent can we follow the patter we do with the `mev_params`; see line 63. If you specify this in that area then the for loop below it will mix the defaults and what was siet

h4ck3rk3y (approved):

src/package_io/parse_input.star
kamilchodola: @h4ck3rk3y Applied :) Tested and works well

PR #326, opened 2023-10-20
fix: genesis validators root had an extra new line

h4ck3rk3y:
was running into the same bug as a few months ago -  https://github.com/flashbots/mev-boost-relay/issues/507


h4ck3rk3y:
image
without the tr

Issue #305, opened 2023-10-18
Add ethereum metrics exporter 


bsamuels453:
https://github.com/crytic/attacknet/issues/5

We need to add the Ethereum metrics explorer to facilitate attacknet health checks

cc @barnabasbusa

barnabasbusa:
cc @samcm

PR #331, opened 2023-10-24
feat(tooling): Add Ethereum Metrics Exporter

samcm:
Adds support for https://github.com/ethpandaops/ethereum-metrics-exporter. An instance for every participant is spun up as EME only supports a single EL/CL pair. Can be enabled on a global or participant level.

Resolves https://github.com/kurtosis-tech/ethereum-package/issues/305


src/ethereum_metrics_exporter/ethereum_metrics_exporter_launcher.star
```diff
@@ -0,0 +1,49 @@
...
+def launch(
+    plan,
+    index,
+    pair_name,
+    ethereum_metrics_exporter_service_name,
+    ethereum_metrics_exporter_image,
+    el_client_context,
+    cl_client_context,
+):
+    listening_port = STARTING_HTTP_PORT_NUMBER + index
```
barnabasbusa, line 18: There is no need to do this. Kurtosis takes care that each service gets its own independent randomized port. So you can safely assign the same port to multiple services.

src/package_io/constants.star
```diff
@@ -31,6 +31,8 @@ GENESIS_VALIDATORS_ROOT_PLACEHOLDER = "GENESIS_VALIDATORS_ROOT_PLACEHOLDER"
 
 DEFAULT_SNOOPER_IMAGE = "ethpandaops/json-rpc-snoop:1.1.0"
 
+DEFAULT_ETHEREUM_METRICS_EXPORTER_IMAGE = "ethpandaops/ethereum-metrics-exporter:0.22.0"
```
barnabasbusa, line 34: You might want to use latest here instead of a tagged version. Just so we don't need to make a new PR for each time this image is updated. We have a static version set for json rpc snoop as that image is not planned to change that frequently.

barnabasbusa:
could you also please add the multi view dashboard?

src/package_io/constants.star
samcm: I don't think we update EME enough for this to be an issue (would also really prefer to not use latest in a testing tool like this)

samcm:
> could you also please add the multi view dashboard?

We don't have a functioning multi view dashboard (the one in Observability Grafana is WIP/broken) 😢. Maybe we can fix that one or create one? Will just take a bit of time

src/ethereum_metrics_exporter/ethereum_metrics_exporter_launcher.star
samcm: Fixed! Tyvm 🔥

barnabasbusa:
Ok lets add this second round around. We should just have this merged in asap so we have a bit more metrics.

barnabasbusa (approved):

h4ck3rk3y (approved):

src/ethereum_metrics_exporter/ethereum_metrics_exporter_launcher.star
```diff
@@ -0,0 +1,54 @@
...
+
+HTTP_PORT_ID = "http"
+METRICS_PORT_NUMBER = 9090
+
+
+def launch(
+    plan,
+    pair_name,
+    ethereum_metrics_exporter_service_name,
+    ethereum_metrics_exporter_image,
```
h4ck3rk3y, line 16: is the intent to make this passable via args later on?

h4ck3rk3y: https://github.com/kurtosis-tech/ethereum-package/issues/334

PR #336, opened 2023-10-25
fix: path based storage - disable for older forks

barnabasbusa:
If capella fork epoch is set, we assume that we want to test an older version of geth.


src/el/geth/geth_launcher.star
```diff
@@ -169,11 +171,14 @@ def get_config(
...
     cmd = [
         "geth",
-        # Disable path based storage scheme for electra fork or when builder image is used
+        # Disable path based storage scheme for electra fork or when builder image or when capella is not 0 is used
         # TODO: REMOVE Once geth default db is path based, and builder rebased
+        # TODO: capella fork epoch check is needed to ensure older versions of geth works.
         "{0}".format(
             "--state.scheme=path"
-            if electra_fork_epoch == None and "--builder" not in extra_params
+            if electra_fork_epoch == None
```
h4ck3rk3y, line 179: its tough to reason here as the comment above is !(a or b or c) while the statement is !a and !b and !c; can we make them align more

h4ck3rk3y (approved):

src/el/geth/geth_launcher.star
h4ck3rk3y: Yep,  I didn't think I would be using De Morgans law; I agree they are the same but squaring the two (code and english) felt tough

PR #343, opened 2023-10-27
fix: builder args incorrectly configured

avalonche:
as new args were added to geth the hardcoded indexes are not configured properly. This makes cmd args for builder more robust.


barnabasbusa (approved):

barnabasbusa:
Thanks for the quick fix, totally missed it!

PR #354, opened 2023-11-06
feat: add prometheus custom configuration for participants

cbermudez97:


src/participant_network.star
```diff
@@ -207,6 +207,11 @@ def launch_participant_network(
             participant.el_extra_env_vars,
         )
 
+        # Add participant el additional prometheus metrics
```
h4ck3rk3y, line 210: i see this is a smaller change than changing every client to add the prometheus config properly - is that why we are doing this?

src/package_io/input_parser.star
```diff
@@ -145,6 +145,10 @@ def input_parser(plan, input_args):
                 ethereum_metrics_exporter_enabled=participant[
                     "ethereum_metrics_exporter_enabled"
                 ],
+                prometheus_config=struct(
```
h4ck3rk3y, line 148: Can you change the readme and network_params.yml to reflect this?

h4ck3rk3y (approved):
@cbermudez97 thank you for the contribution

1. can you tell me what are you trying to do with this?
2. I trust you that the dashboard works with and without args after this change

src/participant_network.star
cbermudez97: Yes, it just passing the prometheus config from a participant to its contexts node metrics info. Since it was a simple association adding that as a param means that every client would have to do this step which is essentially the same. If there is a better place for it I can move it.

h4ck3rk3y: No I am fine keeping it here! Just a bit afraid of people adding more and more things here; but we can address that fear later.

h4ck3rk3y (approved):
Thank you! this looks good to me :-)

PR #357, opened 2023-11-09
feat: Remove overrides

parithosh:
Removes overrides we needed in place to support verkle until transition support was added. The support has now been added.


h4ck3rk3y (approved):

src/el/geth/geth_launcher.star
```diff
@@ -157,8 +157,7 @@ def get_config(
 ):
     # TODO: Remove this once electra fork has path based storage scheme implemented
     if electra_fork_epoch != None:
-        init_datadir_cmd_str = "geth init --cache.preimages --override.prague={0} --datadir={1} {2}".format(
-            final_genesis_timestamp,
+        init_datadir_cmd_str = "geth init --cache.preimages --datadir={0} {1}".format(
```
h4ck3rk3y, line 160: lets wait for the end of the summit!

barnabasbusa:
Could we start having an example dir with all the example yaml files these should work? I'm kinda lost in terms of what geth/lodestar versions I should be testing the current latest petra stuff.

src/el/geth/geth_launcher.star
barnabasbusa: Can we merge this now?

examples/verkle.yaml
```diff
@@ -0,0 +1,14 @@
...
+- el_client_type: geth
+  el_client_image: ethpandaops/geth:transition-post-genesis-21cebf0
+  cl_client_type: lighthouse
+  cl_client_image: ethpandaops/lighthouse:verkle-trees-capella-2ffb8a9
+network_params:
+  deneb_fork_epoch: 500
+  electra_fork_epoch: 1
+  genesis_delay: 100
+launch_additional_services: true
+additional_services: ["el_forkmon", "tx_spammer", "dora"]
```
barnabasbusa, line 12: you probably wanna use a yaml array for this, like 
```
additional_services: 
    - el_forkmon
    - tx_spammer
    - dora
  ```

File: src/participant_network.star [hunk 3]
```diff
@@ -102,7 +102,7 @@ def launch_participant_network(
     # we are running electra - experimental
     elif network_params.electra_fork_epoch != None:
         ethereum_genesis_generator_image = (
-            "ethpandaops/ethereum-genesis-generator:3.0.0-rc.14"
+            "ethpandaops/ethereum-genesis-generator:3.0.0-rc.16"
```
barnabasbusa, line 105: lets override the txfuzz img right here too with the one that is needed for verkle.

src/el/geth/geth_launcher.star
barnabasbusa: @parithosh you sure we can remove the override prague bit?

parithosh: yh cause prague isnt at genesis right, its happening at a defined point in time. So its nothing geth needs to know at genesis, but rather at runtime is enough. You can test it if you like :D

barnabasbusa:
Lets block this PR till we have all ELs working with transition, else they might want to test verkle genesis still.

parithosh:
> Lets block this PR till we have all ELs working with transition, else they might want to test verkle genesis still.

ah i'd make the the case we shouldnt as ELs can use the PR to begin testing and finding bugs in their transition :D
geth is ready and would most likely benefit from this PR

barnabasbusa (approved):

PR #373, opened 2023-11-28
Disable CL

gr8h:


barnabasbusa:
I think this PR could be made a lot more simple. You probably can do the checks if the CL image is set to `null`, and check for that. 

If CL img is set to null, then don't spin it up, else do.

barnabasbusa:
Could you add the aura config into examples/aura.yaml instead of network config?

examples/aura.yaml
```diff
@@ -0,0 +1,106 @@
...
+    el_max_mem: 0
+    bn_min_cpu: 0
+    bn_max_cpu: 0
+    bn_min_mem: 0
+    bn_max_mem: 0
+    v_min_cpu: 0
+    v_max_cpu: 0
+    v_min_mem: 0
+    v_max_mem: 0
+    count: 1
```
barnabasbusa, line 30:
```suggestion

```

If you are using the defaults, then you don't need to redefine them.

examples/aura.yaml
```diff
@@ -0,0 +1,106 @@
...
+      "--config=/nethermind/configs/validator.cfg",
+      "--Init.ChainSpecPath=/nethermind/chainspec/chainspec.json"
+    ]
+    cl_client_type: na
+    cl_client_image: null
+    cl_client_log_level: ""
+    beacon_extra_params: []
+    validator_extra_params: []
+    builder_network_params: null
+    validator_count: null
```
barnabasbusa, line 15:
```suggestion
    validator_count: 0
```
If you set this to 0 it will not bother you with a validator process at all.

examples/aura.yaml
```diff
@@ -0,0 +1,106 @@
+participants:
+  - el_client_type: nethermind
+    el_client_image: nethermind/nethermind:kurtosis
+    el_client_log_level: ""
+    el_extra_params: [
+      "--config=/nethermind/configs/validator.cfg",
+      "--Init.ChainSpecPath=/nethermind/chainspec/chainspec.json"
+    ]
+    cl_client_type: na
```
barnabasbusa, line 9: Is this still needed?

barnabasbusa (changes requested):

barnabasbusa:
Please run `kurtosis lint --format` in order to have no linting errors.

examples/aura.yaml
gr8h: setting `validator_count: 0` for the first participant throws an error `failed to compute sync committee indices: no active validators to compute sync committee from`

h4ck3rk3y:
@gr8h Thanks for the contribution - some notes

1. Can we enrich the README about this behavior
2. While `cl_client_image` being null does work; this feels like magical behavior which is worse when things aren't documented. How about `cl_client_type` being set to null? @barnabasbusa does it have to be the image?

barnabasbusa:
> 2. While `cl_client_image` being null does work; this feels like magical behavior which is worse when things aren't documented. How about `cl_client_type` being set to null? @barnabasbusa does it have to be the image?

Actually cl_client_type being null is probably better, the image being set to null is not logical indeed.

examples/aura.yaml
gr8h: It can be removed, but the node name will be `el-1-nethermind-lighthouse` I guess is the default value.

examples/aura.yaml
```diff
@@ -0,0 +1,39 @@
+participants:
```
h4ck3rk3y, line 1: Lets update the configuration in the readme to talk about the `cl_client_type` being set to null behavior! Sorry for the confusion we go with `cl_client_type` instead of `cl_client_image`!

gr8h:
> > 2. While `cl_client_image` being null does work; this feels like magical behavior which is worse when things aren't documented. How about `cl_client_type` being set to null? @barnabasbusa does it have to be the image?
> 
> Actually cl_client_type being null is probably better, the image being set to null is not logical indeed.

Okay, Make sense.

gr8h:
@barnabasbusa one thing i noticed is that the `el_extra_params` value in the yaml file does not override the values in `nethermind_****.start` file, it only extend it and it duplicate the params, is this an intended behaviour?

barnabasbusa:
>  is this an intended behaviour?

yes it is. 

As its supposed to be extra arguments to a sane default, which would mean that the operation is append, not a find and replace.

barnabasbusa:
Will nethermind be able to take the later values?

gr8h:
> Will nethermind be able to take the later values?

I had to change chainspecs path in the `constants.star`, and removed `"--config=none.cfg"` on nethermind launch script.

PR #382, opened 2023-12-05
fix: peering issue between lighthouse-teku on k8s

h4ck3rk3y:


src/cl/teku/teku_launcher.star
```diff
@@ -257,7 +257,7 @@ def get_config(
         "--p2p-subscribe-all-subnets-enabled=true",
         "--p2p-peer-lower-bound={0}".format(MIN_PEERS),
         "--p2p-advertised-ip=" + PRIVATE_IP_ADDRESS_PLACEHOLDER,
-        "--p2p-discovery-site-local-addresses-enabled",
+        "--p2p-discovery-site-local-addresses-enabled=true",
```
h4ck3rk3y, line 260: this is just me explicitly labelling that flag as true

src/cl/lighthouse/lighthouse_launcher.star
```diff
@@ -188,7 +188,7 @@ def launch(
         port_id=BEACON_HTTP_PORT_ID,
         extract={
             "enr": ".data.enr",
-            "multiaddr": ".data.discovery_addresses[0]",
+            "multiaddr": ".data.p2p_addresses[0]",
```
h4ck3rk3y, line 191: @barnabasbusa this seems like the actual fix; the discovery_address is a udp address while the p2p_address is the tcp address; changing this seems to have fixed the following configuration for me at least

```yaml
participants:
  - el_client_type: geth
    cl_client_type: lighthouse
  - el_client_type: geth
    cl_client_type: teku
  - el_client_type: geth
    cl_client_type: lighthouse
additional_services:
  - dora
```

the teku cli expects TCP in its examples; plus the other outputs of `/eth/v1/node/peers` seems to be last connecting via TCP as well

wonder why this all works on docker; maybe k8s is bottlenecked on bootnodes while docker isnt'?

barnabasbusa (approved):

src/cl/teku/teku_launcher.star
tedim52: Did this actually have any effect? If so, could be worth updating the documentation here - https://docs.teku.consensys.io/reference/cli#p2p-discovery-site-local-addresses-enabled

h4ck3rk3y: this didn't, i was just confused whether its true or false so hardset it

PR #419, opened 2023-12-19
feat: add assertoor to additional toolings

pk910:
Added assertoor testnet testing tool.

There are currently 6 tests included:
* Check for chain stability (finalizing, high voting consensus, no reorgs)
* Check if all clients propose a block
* Validator Lifecycle Test
  Test for the whole validator lifecycle,  includes Deposits, BLSChanges, Exits & Slashings
  This Test takes up to 2 days to run through due to deposit delays & queues. 
* Normal Transaction Test
  Test if normal EOA transactions (type 0/2) are processed by all client pairs (inclusion & submission)
* Blob Transaction Test
  Test if blob transactions (type 3) are processed by all client pairs (inclusion & submission)
* All-Opcodes Transaction Test
  Generates & submits a transaction that triggers every EVM opcode once.
  Checks if the transaction succeeds and no client forks off


pk910:
Ready for first review :)

I'm using this config to run through the tests faster:
```
participants:
  - el_client_type: geth
    el_client_image: ethpandaops/geth:master-5b57727
    cl_client_type: lighthouse
    cl_client_image: ethpandaops/lighthouse:sidecar-inclusion-proof-ee3767c
    count: 1
  - el_client_type: geth
    el_client_image: ethpandaops/geth:master-5b57727
    cl_client_type: teku
    cl_client_image: ethpandaops/teku:master-992b224
    count: 1
network_params:
  deneb_fork_epoch: 1
  genesis_delay: 100
  seconds_per_slot: 3
  num_validator_keys_per_node: 250
launch_additional_services: true
additional_services:
  - dora
  - assertoor
snooper_enabled: true
assertoor_params:
  run_stability_check: true
  run_block_proposal_check: true
  run_lifecycle_test: true
  run_transaction_test: true
  run_blob_transaction_test: true
  run_opcodes_transaction_test: true
```

`run_lifecycle_test` will take quite long to complete.
All other tests are much faster.

barnabasbusa:
Could you please add a new config as a github test, so it is ran overnight? 

How long should one of these tests last? (assuming `run_lifecycle_test: false` )

pk910:
> Could you please add a new config as a github test, so it is ran overnight?

I don't get that. Yea, I think that's actually the plan to run this test later on :)
But it's out of scope for this PR, as it'll be in a separate repository targeting our own runner for it?

> How long should one of these tests last? (assuming run_lifecycle_test: false )

The other tests (except the validator lifecycle one) shouldn't take too long.
It should run through in a few epochs.

The first two "check"-tests are not doing anything, just monitoring the network for about 1-2 epochs.

The transaction related tests shouldn't take longer than a few mins. 
It depends on when all clients got a chance to propose a block with a relevant TX.

barnabasbusa (approved):

PR #449, opened 2024-01-10
fix: dora alternative template

barnabasbusa:


barnabasbusa:
Closing in favor of #452

PR #454, opened 2024-01-10
fix: erigon command arg

barnabasbusa:


src/el/erigon/erigon_launcher.star
```diff
@@ -250,6 +250,8 @@ def get_config(
     if network not in constants.PUBLIC_NETWORKS:
         command_arg = [init_datadir_cmd_str, " ".join(cmd)]
         command_arg_str = " && ".join(command_arg)
```
h4ck3rk3y, line 252: do we need another if condition; can we just take this one out at the end?

h4ck3rk3y (approved):

PR #466, opened 2024-01-17
feat: Add Xatu Sentry

samcm:
Adds support for [Xatu](https://github.com/ethpandaops/xatu) in Sentry mode. Similarly to Ethereum Metrics Exporter, Xatu Sentry can be enabled globally,  or per client pair. 

Note: This PR only enables "Sentry". Xatu Server/Clickhouse etc will need to run outside of Kurtosis for this to be useful.


barnabasbusa (approved):

h4ck3rk3y:
Thanks for the contirbution Sam! :-)

PR #467, opened 2024-01-17
fix: Update nethermind to expose host on 0.0.0.0

barnabasbusa:
Fixing this: https://github.com/NethermindEth/nethermind/pull/6528/


barnabasbusa:
Once there is a new release for nethermind, we can default back to using :latest

parithosh (approved):

Issue #372, opened 2023-11-28
add support for blockscout


h4ck3rk3y:

franjoespejo:
Really important for us, specially if the https://docs.blockscout.com/for-users/verifying-a-smart-contract module is enabled

barnabasbusa:
@franjoespejo would you consider contributing this tooling ?

franjoespejo:
Thanks @barnabasbusa! would love but we cannot commit to it atm. We are arranging things to contribute very soon, would take this ticket if is still open

igor-ritual:
Working rough draft: https://github.com/kurtosis-tech/ethereum-package/compare/main...igor-ritual:ethereum-package:blockscout

barnabasbusa:
I'd recommend not having this package enabled by default, but make it an optionally configurable package. 

Feel free to open a PR when you feel like its ready for review.

barnabasbusa:
Would you please add mem/cpu requests/limits to the application? Similar to https://github.com/kurtosis-tech/ethereum-package/pull/396

barnabasbusa:
This is now merged with #481

PR #479, opened 2024-01-30
feat: blockscout with verif service, WIP

franjoespejo:
solves #372


barnabasbusa:
can you format it with `kurtosis lint --format` ? 

Also I wouldn't set blockscout as default (you had it as a comment)

PR #481, opened 2024-01-31
feat: blockscout support with sc verification

franjoespejo:
addresses #372


franjoespejo:
@barnabasbusa pls note:
- blockscout execution block explorer not activated by default
- up to date with latest changes in main
- resources limit values are just a guess
- not sure if its ok to return the 'blockscout verification url' from the main function so can be used in the diva-package that calls this eth-package

barnabasbusa:
Could you add blockscout to a couple of the tests ? 

- `.github/tests/mix-with-tools.yaml`
- `.github/tests/mix-with-tools-mev.yaml`

Do you see a value in enabling persistent storage for this tool (if users asks for persistence) ?

franjoespejo:
right, will do test and persistence!

franjoespejo:
I also returned the url for the verification sc service, pls check, I can rm if needed

src/blockscout/blockscout_launcher.star
```diff
@@ -0,0 +1,131 @@
...
+}
+
+
+def launch_blockscout(plan, el_client_contexts, persistent):
+    postgres_output = postgres.run(
+        plan,
+        service_name="{}-postgres".format(SERVICE_NAME_BLOCKSCOUT),
+        database="blockscout",
+        extra_configs=["max_connections=1000"],
+        persistent=persistent,
```
barnabasbusa, line 47: This won't enable persistence the way you expect it. 

It needs to be defined like this: 

https://github.com/kurtosis-tech/ethereum-package/blob/main/src/el/geth/geth_launcher.star#L333

So its the files parameter that you want to enable persistence on. And mount the pvc to whatever datadir you want to keep persistence of.

barnabasbusa:
This probably has more to do with the image than with kurtosis config, so lets merge this.

barnabasbusa (approved):

PR #491, opened 2024-02-09
feat!: add node selectors features

barnabasbusa:


barnabasbusa:
Requires kt version  0.86.17

.github/tests/node-selectors.yaml
```diff
@@ -0,0 +1,13 @@
...
+    cl_client_type: teku
+    cl_split_mode_enabled: true
+    node_selectors: {
+      "kubernetes.io/hostname": testing-1,
+    }
+  - el_client_type: reth
+    cl_client_type: teku
+    cl_split_mode_enabled: true
+global_node_selectors: {
+  " kubernetes.io/hostname": testing-2,
```
h4ck3rk3y, line 12: extra space

h4ck3rk3y: seems like this test isn't running? otherwise why isn't it failing

h4ck3rk3y: this should be moved into Circle and run as a k8s test!

h4ck3rk3y (approved):

.github/tests/node-selectors.yaml
h4ck3rk3y, line 12:
```suggestion
  "kubernetes.io/hostname": testing-2,
```

barnabasbusa: If we move it to Circleci, then its going to continuously fail. I rather leave it as is, and let it not fail.

PR #494, opened 2024-02-12
fix: holesky config

parithosh:
holesky already went through dencun


barnabasbusa:
This will still not work, as by default the deneb fork is set to be 500.

h4ck3rk3y:
@parithosh @barnabasbusa  should we close this?

barnabasbusa:
Fixed by #500

PR #497, opened 2024-02-20
feat!: separate validator clients from CL clients

eth2353:
Separates the validator clients more cleanly from the CL clients. This then allows the use of different combinations of CL/VC clients, e.g. Teku VC with Lodestar CL. The only VC that doesn't work with different beacon nodes is Prysm at the moment.

The `use_separate_validator_client` flag defaults to false for CL clients that can run validators in the same process as the CL (preserving the way the `cl_split_mode_enabled` worked before this PR).

I believe this can be quite useful to test different VC<->CL combinations for compatibility.


eth2353:
I am unable to add a reviewer to this PR myself.

@barnabasbusa I believe you would be the best person to review this since you introduced the `cl_split_mode_enabled` flag, would you mind taking a look at this? Thanks!

barnabasbusa:
I'll try to review this hopefully by the end of the week.

barnabasbusa:
Sorry I still haven't had time to deeply look into this PR as it is touching quite a bit, I'll make sure to take a proper look before we can merge this in. 
Can you please update all the tests, especially the ones that define a prysm image, as it looks like all those tests will fail. 
I'll take a look if there would be a way for us to run our nightly tests on this PR.

eth2353:
I suppose these may look like rather large changes but they really aren't _that_ large. It's mostly refactoring/relocating existing code. But I understand you want to take a good look at this, take all the time you need.

I updated the prysm CL image definitions in the test configs, you're right, those configs with the previously used 2 image definitions would have failed.

I ran the assertoor, ephemery, mev, split-nimbus, split-teku examples locally (discovered one more thing I overlooked when a network participant has 0 keystores).

Next, I ran all CL client combinations with all valid `use_separate_validator_client` values. I also tried out some variations with `validator_count: 0`, everything worked as expected.

One more thing, I added an example config to the tests folder which shows an example of mixed CL / validator client usage. Hope that's okay, it's meant to illustrate the usecase, but feel free to remove it if you want.

README.md
```diff
@@ -203,27 +203,40 @@ participants:
...
   cl_client_log_level: ""
 
-  # A list of optional extra params that will be passed to the CL to run separate Beacon and validator nodes
-  # Only possible for nimbus or teku
-  # Please note that in order to get it to work with Nimbus, you have to use `ethpandaops/nimbus:unstable` as the image (default upstream image does not yet support this out of the box)
-  # Defaults to false
-  cl_split_mode_enabled: false
+  # Whether to use a separate validator client attached to the CL client.
+  # Defaults to false for clients that can run both in one process (Teku, Nimbus)
+  use_separate_validator_client: true/false
```
barnabasbusa, line 224: in the readme please use the default value

barnabasbusa (approved):

PR #502, opened 2024-02-28
feat: add keymanager to all validator processes

barnabasbusa:


src/validator_client/lighthouse.star
```diff
@@ -74,6 +81,12 @@ def get_config(
         + el_client_context.client_name,
     ]
 
+    is_verkle = (
+        "verkle" in network or electra_fork_epoch != None
+    )  # TODO: remove when deneb rebase is done
+    if not is_verkle:
```
h4ck3rk3y, line 87: just do `if "verkle" in netowrk or elecrtra_fork_epoch !=None`

static_files/keymanager/run.sh
```diff
@@ -0,0 +1,7 @@
+# To run this script, you need to have openssl installed on your machine
```
h4ck3rk3y, line 1: maybe this file can be renamed to "generate_certs.sh"

h4ck3rk3y (approved):

PR #503, opened 2024-02-29
fix: use the cl as the default validator image if none are defined

barnabasbusa:
This PR ensures that the validator image will be the same as CL image (if one is defined) and only fallback to the default validator image if there is no validator image nor beacon image is defined.


barnabasbusa:
Gotta merge this: https://github.com/ethpandaops/eth-client-docker-image-builder/pull/134 to pass all the CI jobs.

Have to rename all references from nimbus to nimbus-eth2 where the panda images are used.

pk910 (approved):

PR #529, opened 2024-03-28
fix: disable static peers

barnabasbusa:


barnabasbusa:
Peering doesn't seem to work for nethermind

parithosh (approved):

PR #531, opened 2024-03-29
fix: beaconchain explorer

barnabasbusa:


TobiWo (approved):
Looks good to me although I would create new variables and not slice directly because it is more readable if you slice into a well named variable :slightly_smiling_face: .

barnabasbusa:
Fixed. @TobiWo

parithosh (approved):

PR #547, opened 2024-04-08
fix: change rpc to http port ids

barnabasbusa:


barnabasbusa:
Closing in favor of #548

PR #552, opened 2024-04-10
chore: added a doc string for better args

h4ck3rk3y:
This should improve the doc string for the arguments provided to Kurtosis; I am not sure how to test this easily in the UI before merging this but I ran the linter on this and the string is valid

image

this is how it currently looks like; after this gets added the args will have the string description given in this PR

image

this pure YAML/JSON field will still expect something like

```
args:
   participants:
```


h4ck3rk3y:
image

this is how something with a description looks like

barnabasbusa (approved):

PR #567, opened 2024-04-16
feat: use new rpc snooper from `ethpandaops/rpc-snooper`

pk910:
This PR replaces the existing RPC snooper with a new one.

The existing snooper unfortunately causes several problems when intercepting CL<>VC connections.
This is due to lack of proper support for the event stream calls. 
These event stream calls are currently timing out after roughly a minute and do not serve any events.

The new snooper properly handles event stream calls and adds the invoked events to the log output.
Besides of that it supports all the common snooper functionality.
It logs prettified JSON and hex encoded SSZ bodies as usual.

**This is a WIP PR.**


pk910:
Test with all CL<>VC pairs shows same result as without snooper:
https://github.com/ethpandaops/assertoor-test/actions/runs/8712937458

Test with all EL<>CL fully works:
https://github.com/ethpandaops/assertoor-test/actions/runs/8713231136

both workflow runs were executed with this branch.

barnabasbusa (approved):

PR #582, opened 2024-04-29
fix!: upcoming file path change in kurtosis upstream

barnabasbusa:
This PR fixes the compatibility issues with the latest kurtosis release, and requires engine version 0.89.0 at least.


h4ck3rk3y (approved):
LGTM

PR #584, opened 2024-04-30
feat: allow setting exit ip address

h4ck3rk3y:


h4ck3rk3y:
image

barnabasbusa:
can you paste an example of enr too? 

Does it still work on your local machine if you set it to auto? If so, maybe would make sense to use that by default?

barnabasbusa (approved):

PR #612, opened 2024-05-14
fix: tune Besu options to work with tx_spammer

fab-10:
At startup Besu does not return a good gas price estimation (WIP to fix this) and the default number of accepted connection is not enough to sustain the tx_spammer, so this PR tune some options to allow to use the tx_spammer wth Besu.


fab-10:
@pk910 could you please review?

.github/tests/besu-all.yaml
```diff
@@ -1,14 +1,15 @@
...
     cl_type: nimbus
   - el_type: besu
     cl_type: lighthouse
   - el_type: besu
     cl_type: lodestar
-additional_services: []
+  - el_type: besu
+    cl_type: grandine
+additional_services:
+  - tx_spammer
```
barnabasbusa, line 15:
```suggestion
additional_services: []
```

These tests are ran overnight. There is no need to test tx_spammer here. You can test that locally.

barnabasbusa (approved):

pk910:
Yea, our testing workflow is currently broken for branches that are not part of this repository.
Right now it tries to find branch `tune-besu-for-tx-spammer` within this repository, but it's part of your fork, so it can't find it.
We'll get that fixed :)

pk910:
PR workflow issue with assertoor got fixed with #616 
Sorry for the delay ;)

PR #617, opened 2024-05-15
chore: see if sed solves the issue

h4ck3rk3y:


barnabasbusa:
Could you got do github checkout PR, then use that to run kurtosis run . with assertoor test?

PR #634, opened 2024-05-23
fix: re-add images to labels

barnabasbusa:


skylenet (approved):

prestonvanloon:
Could this be causing https://github.com/kurtosis-tech/kurtosis/issues/2460 ?

Seems like `@` is not acceptable in a k8s label.

PR #640, opened 2024-05-26
fix: Vc count logic

0xTylerHolmes:
I ran into a few index out of bounds and divide by zero issues. I fixed those with what I expect is the right method. One thing I would like you to check is my change on the [participant_network.star](https://github.com/0xTylerHolmes/ethereum-package/blob/061dbec8694e78dfeb2cf9c48d351202866fd6e6/src/participant_network.star#L397)

I added a check on the `vc_context` assignment for the case where `vc_count == 0`. We would get an index out of bounds panic so I just set it to `None`. I'm unsure if there are downstream consequences but everything appears to be working on my end with that.

### DIFF
--- a/src/cl/cl_launcher.star
+++ b/src/cl/cl_launcher.star
@@ -119,7 +119,7 @@ def launch(
 
         cl_service_name = "cl-{0}-{1}-{2}".format(index_str, cl_type, el_type)
         new_cl_node_validator_keystores = None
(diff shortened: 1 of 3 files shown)


barnabasbusa:
can you please run `kurtosis lint --format` to lint this PR?

barnabasbusa:
`vc_context = None` should work fine, if we have no vc client set.

PR #646, opened 2024-05-27
feat: Adding arbitrary contract definition

parithosh:


src/package_io/constants.star
```diff
@@ -88,7 +88,7 @@ EIP7594_FORK_VERSION = "0x70000038"
 
 ETHEREUM_GENESIS_GENERATOR = struct(
     capella_genesis="ethpandaops/ethereum-genesis-generator:2.0.12",  # Deprecated (no support for minimal config)
-    deneb_genesis="ethpandaops/ethereum-genesis-generator:3.1.6",  # Default
+    deneb_genesis="parithoshj/ethereum-genesis-generator:allow-prefunded-contract-def-v1",  # Default
```
parithosh, line 91: Needs a release, and to be changed

network_params.yaml
```diff
@@ -80,6 +80,7 @@ network_params:
   samples_per_slot: 8
   custody_requirement: 1
   target_number_of_peers: 70
+  additional_preloaded_contracts: '{"0x123463a4B065722E99115D6c222f267d9cABb524": {"balance": "1ETH","code": "0x1234","storage": {},"nonce": 0,"secretKey": "0x"}}'
```
barnabasbusa, line 83: Does this actually do anything? Why not leave it `{}`?

parithosh: Adds an example so people can see how to use it, but yeah default can be {} if needed

barnabasbusa: Would add the example in the github tests

src/package_io/constants.star
barnabasbusa, line 91:
```suggestion
    deneb_genesis="ethpandaops/ethereum-genesis-generator:3.2.0",  # Default
```

network_params.yaml
parithosh: added an example in github tests and removed from here

.github/tests/custom-contracts.yaml
```diff
@@ -0,0 +1,18 @@
+participants:
+  - el_type: geth
+    cl_type: teku
+  - el_type: nethermind
+    cl_type: prysm
+    cl_image: gcr.io/prysmaticlabs/prysm/beacon-chain:stable
```
barnabasbusa, line 6:
```suggestion
```

barnabasbusa (approved):

PR #685, opened 2024-06-20
fix: blockscout bad return

barnabasbusa:


barnabasbusa:
Fixing: https://github.com/ethpandaops/optimism-package/issues/39

pk910 (approved):

PR #700, opened 2024-06-28
fix: ephemery genesis loader

pk910:
update ephemery genesis loader to use latest `network-config.tar.gz` from ephemery genesis repository

### DIFF
--- a/src/network_launcher/ephemery.star
+++ b/src/network_launcher/ephemery.star
@@ -9,8 +9,11 @@ def launch(plan, prague_time):
         name="fetch-ephemery-genesis-data",
         description="Creating network configs",
         run="mkdir -p /network-configs/ && \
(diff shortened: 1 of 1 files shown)


barnabasbusa (approved):
lgtm

barnabasbusa:
There is one more file that has to be added. Thats not present in eth-clients repo either

Prysm needs to read in the bootstrap nodes in a yaml format instead of a txt format. 

https://github.com/ethpandaops/ansible-collection-general/blob/5606de4c6ab7e3f69f6b27939f5c9062d83e126f/roles/ethereum_post_network_setup/tasks/main.yaml#L42

pk910:
fixed :)

barnabasbusa (approved):

Issue #713, opened 2024-07-09
start local testnet failed


cpucorecore:
i followd the link:https://github.com/sigp/lighthouse/tree/unstable/scripts/local_testnet to start a ethereum local testnet but got the error below:

```txt
INFO[2024-07-09T11:10:49+08:00] Compressing package 'github.com/ethpandaops/ethereum-package' at '/Users/sky/ethereum-package' for upload 
INFO[2024-07-09T11:10:49+08:00] Uploading and executing package 'github.com/ethpandaops/ethereum-package' 
There was an error interpreting Starlark code 
Evaluation error: Cannot construct 'run_sh' from the provided arguments.
	Caused by: run_sh: unexpected keyword argument "name"
	at [github.com/ethpandaops/ethereum-package/main.star:128:55]: run
	at [github.com/ethpandaops/ethereum-package/src/participant_network.star:90:78]: launch_participant_network
	at [github.com/ethpandaops/ethereum-package/src/prelaunch_data_generator/el_cl_genesis/el_cl_genesis_generator.star:49:26]: generate_el_cl_genesis_data
	at [0:0]: run_sh

Error encountered running Starlark code.
```

barnabasbusa:
Can you please update your kurtosis engine and try again?

hangleang:
I got below error instead, when I tried to start a localnet with a local docker image. It's working fine with remote kurtosis configs.

```
There was an error interpreting Starlark code 
Evaluation error: list index 1 out of range [-1:0]
	at [github.com/ethpandaops/ethereum-package/main.star:140:55]: run
	at [github.com/ethpandaops/ethereum-package/src/participant_network.star:122:33]: launch_participant_network
	at [github.com/ethpandaops/ethereum-package/src/network_launcher/devnet.star:9:44]: launch
	at [github.com/ethpandaops/ethereum-package/src/shared_utils/shared_utils.star:213:65]: calculate_devnet_url

Error encountered running Starlark code.
```

**Environments:**
Kurtosis: v1.0.0
Docker: v25.0.1
OS: MacOS

hangleang:
> I got below error instead, when I tried to start a localnet with a local docker image. It's working fine with remote kurtosis configs.
> 
> ```
> There was an error interpreting Starlark code 
> Evaluation error: list index 1 out of range [-1:0]
> 	at [github.com/ethpandaops/ethereum-package/main.star:140:55]: run
[...]

PR #714, opened 2024-07-09
fix: el_cl_genesis_generator.star

cpucorecore:
fix issue #713


barnabasbusa:
I don't believe this PR is needed if you update your kurtosis engine.

cpucorecore:
> I don't believe this PR is needed if you update your kurtosis engine.

yes, i update kurtosis from 0.89.11 to 1.0.0, it works

PR #725, opened 2024-07-19
fix: correctly apply extra params if builder is enabled

nflaig:
I noticed it is not possible to run mixed vc<>bn clients if builder / mev is enabled as it applies extra flags from other clients in `vc_extra_params` as the params are applied based on `cl_type` but it should be done based on `vc_type`.

I tested this using the [following config](https://gist.github.com/nflaig/949c8b5101127961023d89de9c206b04) with both `mev_type: mock` and `mev_type: flashbots`.


barnabasbusa:
Good catch!

barnabasbusa (approved):

PR #726, opened 2024-07-26
feat: add genesis_gaslimit param configuration

jhkimqd:
This PR adds `genesis_gas_limit` parameter within `network_params.yaml` to make the parameter configurable using Kurtosis.

It seems like [ethereum-genesis-generator](https://github.com/ethpandaops/ethereum-genesis-generator), which is being used in this package already has the capability to change `genesis_gas_limit` by changing `export GENESIS_GASLIMIT="${GENESIS_GASLIMIT:-30000000}"` within [defaults.env](https://github.com/ethpandaops/ethereum-genesis-generator/blob/master/defaults/defaults.env#L27). This feature didn't seem like it was directly transferred to the `ehtereum-package`.

Tested locally with the `main` branch and changing the `genesis_gas_limit` value to `50000000`.
Screenshot from 2024-07-26 11-26-12


barnabasbusa:
please add the extra flag into the readme too, and default to 30M gas limit instead of 25.

README.md
```diff
@@ -464,6 +464,9 @@ network_params:
   # How long you want the network to wait before starting up
   genesis_delay: 20
 
+  # The gas limit of the network set at genesis
+  genesis_gas_limit: 30000000
```
barnabasbusa, line 468: Could you please use `genesis_gaslimit` instead of `genesis_gas_limit`everywhere?

```suggestion
  genesis_gaslimit: 30000000
```
This would match the naming convention we have introduced earlier.

barnabasbusa (approved):

PR #730, opened 2024-08-04
feat: lodestar persists invalid ssz objects by default

nflaig:
Would be good to have invalid ssz objects persisted by default in case there is a consensus bug or other issue. There is not much downside to this besides slight storage increase but since states / blocks are generally small on devnets this shouldn't matter, also we prune objects after some time (15 days).


barnabasbusa:
Is this flag an older flag? Is it present in all the devnet environments? (pectra/peerdas branches)

barnabasbusa (approved):

nflaig:
> Is this flag an older flag? Is it present in all the devnet environments? (pectra/peerdas branches)

Yes, the flag is quite old, don't know the exact release but must be more than a year ago

PR #735, opened 2024-08-06
chore: Update blobscan docker images

PabloCastellano:
These new images are much more optimized and are ten times lighter (https://github.com/Blobscan/blobscan/pull/390)


barnabasbusa:
LGTM

barnabasbusa (approved):

PR #746, opened 2024-08-19
Change default churn quotient

parithosh:
This change is required to reduce the staked ether needed in local testnets to test consolidations.


parithosh:
Decided against changing the default and deviating from mainnet values

PR #754, opened 2024-08-28
fix: remove epoch checker for goomy

barnabasbusa:


src/goomy_blob/goomy_blob.star
```diff
@@ -58,16 +58,6 @@ def get_config(
         cmd=[
             " && ".join(
                 [
```
pk910, line 60: we can remove this && concatenation too :)

pk910 (approved):

PR #769, opened 2024-09-18
Init

whereasjovially:
1. Updated network.params.yaml 
2. Added a new README.md


network_params.yaml
```diff
@@ -59,7 +59,7 @@ network_params:
   network_id: "3151908"
   deposit_contract_address: "0x4242424242424242424242424242424242424242"
   seconds_per_slot: 12
-  num_validator_keys_per_node: 64
+  num_validator_keys_per_node: 32
```
Nomad-Free-Talent, line 62: Just curious, what is the reason of this update?

Nomad-Free-Talent (commented):
Can we update the network_params.yaml file to only start the required containers for the first stage?

barnabasbusa:
Could you please explain why you created a new readme? 
What is the point of enabling certain services as default?

PR #781, opened 2024-09-30
docs: update README grafana example config

samlaf:
Think the prometheus_grafana services are off by default, since I had to modify the config this way to get grafana and prometheus to start.


barnabasbusa (approved):

samlaf:
@barnabasbusa some docker rate limit problems with ci. Prob just rerunning will pass?

PR #786, opened 2024-10-02
feat: add rbuilder remove old geth builder - flashbots

barnabasbusa:


barnabasbusa:
BUG!! Still broken: if you set first participant to be reth, it will use the builder image. BUG !!

skylenet (approved):

PR #788, opened 2024-10-03
feat: add supernode label to every container

barnabasbusa:


src/cl/lodestar/lodestar_launcher.star
```diff
@@ -328,11 +329,12 @@ def get_beacon_config(
...
-            constants.CL_TYPE.lodestar,
-            constants.CLIENT_TYPES.cl,
-            participant.cl_image,
-            el_context.client_name,
-            participant.cl_extra_labels,
+            client=constants.CL_TYPE.lodestar,
+            client_type=constants.CLIENT_TYPES.cl,
+            image=participant.cl_image,
+            connected_client=el_context.client_name,
+            extra_labels=participant.cl_extra_labels,
```
parithosh, line 336: if the `supernode` bit is just used for labels, why not just concat the `extra_labels` with the `supernode` one at the input_parser (or similar) level and call it a day? then you don't have to pipe it through everywhere

barnabasbusa: in the end its the same thing, as the label maker will merge all these together

parithosh: Yeah but the advantage is that the next time you do such a PR, it'll be easier to add and quicker to spot an issue :D

Instead of refactoring later might as well do it now?

parithosh: unless you plan on using the supernode label some other way too?

barnabasbusa: I mean this is a label thats gonna be used always going forward. Adding this to extra_labels instead of as a standard label, would be like always appending something to cl_extra_args instead of hardcoding it. Sure it could be done, but thats not what extra labels namespace should be used. I feel like extra_labels should be used for only labels that are passed during runtime. Otherwise you will find labels in the extra labels field that you were not expecting

parithosh (approved):

src/cl/lodestar/lodestar_launcher.star
parithosh: i think it doesnt really make a diff as a label from prometheus' view is the same and doesnt really matter, but i get your point that the supernode concept is canonical and worth preserving explicitly. fyi the name might change in the future :D 


approved PR, good to merge from me

barnabasbusa: The prometheus label is a different thing tho. This is a docker label. - this could be used for future chaos targetting easily.

PR #793, opened 2024-10-07
chore: prysm - set stable image as default

barnabasbusa:


pk910 (approved):

skylenet:
Due to https://github.com/ethpandaops/ethereum-package/pull/802 , where we added the `http-` flags, we need to wait for them to be available in the `stable` release. 

Can be verified with `docker run --rm -it gcr.io/prysmaticlabs/prysm/beacon-chain:stable --help | grep http-port`

PR #798, opened 2024-10-09
feat: add support for fulu/osaka fusaka

barnabasbusa:


barnabasbusa:
Can confirm that this PR will not break any client.

src/dora/dora_launcher.star
```diff
@@ -123,7 +123,7 @@ def get_config(
     if dora_params.image != "":
         IMAGE_NAME = dora_params.image
     else:
-        IMAGE_NAME = "ethpandaops/dora:latest"
+        IMAGE_NAME = "ethpandaops/dora:master"
```
pk910, line 126: I'd prefer leaving this as `ethpandaops/dora:latest` by default.
That's the stable image that should be used for anyone not working on devnet 4, using kurtosis for research or other deneb related testing. 
They should not be affected by bugs introduced by unstable master builds.

Maybe add a condition, if electra is to be activated, use master,  otherwise use latest?

pk910 (approved):

PR #804, opened 2024-10-10
feat: add checkpoint sync capabilities to nimbus

barnabasbusa:


src/cl/cl_launcher.star
```diff
@@ -146,6 +146,26 @@ def launch(
...
+            if checkpoint_sync_url == "":
+                if (
+                    network_params.network in constants.PUBLIC_NETWORKS
+                    or network_params.network == constants.NETWORK_NAME.ephemery
+                ):
+                    plan.print(
+                        "we got here because network is public, {0}".format(
+                            network_params.network
+                        )
+                    )
```
pk910, line 160: looks like some leftover from debugging? ;)

src/cl/nimbus/nimbus_launcher.star
```diff
@@ -22,7 +22,7 @@ BEACON_MIN_CPU = 50
 BEACON_MIN_MEMORY = 256
 
 DEFAULT_BEACON_IMAGE_ENTRYPOINT = ["nimbus_beacon_node"]
-
+BEACON_NODE_ENTRYPOINT = "/home/user/nimbus-eth2/build/nimbus_beacon_node"
```
pk910, line 25: It looks very wrong hardcoding this path here as its likely gonna change when nimbus touches their build system.
But yea,  don't really see a better solution :/
Do our ethpandaops nimbus images use the same entrypoint?

src/cl/cl_launcher.star
barnabasbusa: yeah :D

src/cl/nimbus/nimbus_launcher.star
barnabasbusa: yes, works with our images too. 

We can always change it if they decide to change their entrypoint. For now I think this is our "only" reasonable solution.

pk910 (approved):

PR #808, opened 2024-10-23
feat: add sticky clients for devnets

barnabasbusa:


barnabasbusa:
We should have this issue resolved with bootnoder

PR #819, opened 2024-10-30
fix: docker login to separate action yaml

skylenet:
- Only run docker login if credentials are available.
- Print rate limits.


skylenet:
Ci jobs on main branch were updated to use kurtosis `1.3.1` , so that we don't have the docker auth bug. CI tests pass again now.

parithosh (approved):

PR #827, opened 2024-11-05
feat: add vero `vc_type`

eth2353:
Adds the Vero validator client as a supported `vc_type`.

The Vero VC doesn't handle keys and requires a remote signer to be used (`use_remote_signer: true`). It can handle the minimal preset, as well as custom `seconds_per_slot` values.

I've added the "Vero - Detailed" Grafana dashboard. I'm not sure if it's a good fit for this repo, let me know if you'd like me to remove it.


eth2353:
Grafana dashboard modified to show the job.

> possibly have an option to enable overview of all validator processes at the same time.

This isn't that easy to do in this dashboard, it's really quite instance-specific.

parithosh (approved):

PR #828, opened 2024-11-05
fix: rbuilder upstream repo with reth-rbuilder binary

barnabasbusa:
Depends on https://github.com/ethpandaops/eth-client-docker-image-builder/pull/208
Depends on https://github.com/flashbots/rbuilder/pull/237


barnabasbusa:
It looks like the current implementation uses reth 1.0.6 which does not seem to have flashbots in the http.api available. 

error: Invalid value 'admin,net,eth,web3,debug,trace,flashbots' for --http.api <HTTP_API>: Matching variant not found.
    [possible values: admin,debug,eth,net,trace,txpool,web3,rpc,reth,ots]

parithosh (approved):

PR #833, opened 2024-11-22
feat: add support for pull through cache

barnabasbusa:


barnabasbusa:
image

when running with cache enabled ^

barnabasbusa:
The only two images that are not going to be pulled through the cache are images that are default assigned by kurtosis - [plan.run_python](https://docs.kurtosis.com/api-reference/starlark-reference/plan/#run_python) - python:3.11-alpine and [plan.run_sh](https://docs.kurtosis.com/api-reference/starlark-reference/plan/#run_sh)  - badouralix/curl-jq.

However, these images are tiny, and there is only 2 of them, so these shouldn't need to be pulled through a cache anyway.

Scutua (approved):

Scutua (approved):

PR #838, opened 2024-11-28
feat: Add-blockscout_params

### LINKED ISSUE #837: Add blockscout configuration
Currently, we cannot choose image of blockscout and blockscout smart contract verifier. it would be better if we can define the image used here.

zt-9:
fix #837 
the default blockscout image version does not work on my machine. 
so i added the blockscout_params for blockscout service.  now user can customize the image versions they want to use.
it contains two params:
```yaml
blockscout_params:
  # blockscout docker image to use
  # Defaults to the latest image
  image: "blockscout/blockscout:latest"
  # blockscout smart contract verifier image to use
  # Defaults to ghcr.io/blockscout/smart-contract-verifier:v1.9.0
  verif_image: "ghcr.io/blockscout/smart-contract-verifier:v1.9.0"
```

i already tested it on my machine and it worke well.


barnabasbusa:
The newest version of blockscout also has a frontend image. 
Also currently their latest image is incompatible with kurtosis.

zt-9:
> The newest version of blockscout also has a frontend image. Also currently their latest image is incompatible with kurtosis.

updated to "blockscout/blockscout:6.8.0"

barnabasbusa (approved):

barnabasbusa:
could you please run a linter?

`kurtosis lint --format`

Ibraimov23:
image
He asks for my wallet details. Does this affect the solution to my problem?

PR #850, opened 2024-12-06
feat: add spamoor

barnabasbusa:


README.md
```diff
@@ -927,6 +927,19 @@ checkpoint_sync_enabled: false
...
+# Spamoor params
+spamoor_params:
+  # The image to use for spamoor
+  image: ethpandaops/spamoor:latest
+  # Throughput of spamoor
+  throughput: 1000
+  # Max pending transactions for spamoor
+  max_pending: 1000
+  # Max wallets for spamoor
+  max_wallets: 1000
```
pk910, line 939: 500 should be enough :)
funding these wallets is a sequential process that consumes time on startup,   we probably don't need more than 500 wallets in parallel ever (max 100 pending tx are usually accepted for each wallet,  so 500 wallets are enough for 50000 pending tx)

pk910 (approved):

PR #858, opened 2024-12-13
fix: add fulu overrides automatically for assertoor and dora if fulu is active

barnabasbusa:


src/assertoor/assertoor_launcher.star
```diff
@@ -119,7 +119,11 @@ def get_config(
         ASSERTOOR_CONFIG_FILENAME,
     )
 
-    IMAGE_NAME = assertoor_params.image
+    if assertoor_params.image == constants.DEFAULT_ASSERTOOR_IMAGE:
+        if network_params.fulu_fork_epoch < constants.FULU_FORK_EPOCH:
+            IMAGE_NAME = "ethpandaops/assertoor:fulu-support"
+    else:
+        IMAGE_NAME = assertoor_params.image
```
pk910, line 126: IMAGE_NAME can be unset here when using the default image and default fuku_fork_epoch

`IMAGE_NAME = assertoor_params.image`  missing for inner if-else clause

pk910 (approved):

PR #880, opened 2025-01-26
feat: Adding support for EIP-7732 and EIP-7805

parithosh:
- Awaiting a genesis generator update to consume these values


terencechain (approved):
thank you!

pk910 (approved):

PR #905, opened 2025-02-18
fix: provide `--network-custom-config-path` to Vero

eth2353:
As of `v0.9.0` Vero [requires a network to be specified as a CLI argument](https://github.com/serenita-org/vero/releases/tag/v0.9.0). This PR adds that to the Kurtosis definition for the Vero VC service.

(In case you're wondering - how did Vero not have a `--network` flag before? It used to load all the spec values from the connected beacon node(s) and would then dynamically create all necessary SSZ classes.)

I tested this manually with the `.github/tests/vero-all.yaml` config and have been using this version for the last few months while preparing for Pectra.

### DIFF
--- a/src/vc/vc_launcher.star
+++ b/src/vc/vc_launcher.star
@@ -159,6 +159,7 @@ def launch(
             fail("vero VC doesn't support the Keymanager API")
         config = vero.get_config(
             participant=participant,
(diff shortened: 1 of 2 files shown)


barnabasbusa (approved):

barnabasbusa:
This is nice, any reason you can’t rely on the beacon node to load this info from tho?

PR #915, opened 2025-03-10
fix: lighthouse minimal image

barnabasbusa:


barnabasbusa:
Not super nice solution, but till lh pushes their minimal fix to stable, this will do

parithosh (approved):

PR #924, opened 2025-03-19
fix: prometheus shouldnt use latest

barnabasbusa:
Prometheus for some reason overrides their bugfix releases for older versions of major prom versions with `latest`, breaking everything for everyone else.


parithosh (approved):

barnabasbusa:
We can revert this change once https://github.com/prometheus/prometheus/issues/16238 is fixed.

PR #948, opened 2025-04-07
feat: add spammor_blob wrapper v1 to activate with fulu

barnabasbusa:


parithosh:
Do we have any legacy ethereum-genesis-generators in use now? otherwise the default on `osaka_time` or `prague_time` might not be set. As long as that's true - should be fine to include these vars

parithosh (approved):

barnabasbusa:
We do not, we use 4.x.x now everywhere. Which is also able to do legacy genesises.

src/spamoor_blob/spamoor_blob.star
```diff
@@ -38,6 +46,14 @@ def get_config(
         ),
     ]
 
+    if (
+        "peerdas" in network_params.network
+        or network_params.fulu_fork_epoch != constants.FAR_FUTURE_EPOCH
+    ):
+        image = "ethpandaops/spamoor:blob-v1"
```
jimmygchen, line 53: @barnabasbusa I can't override this image anymore, would it be possible to add `image = spamoor_params.image` to after this to allow override?

PR #1007, opened 2025-05-08
fix: lighthouse to be able to run alone

barnabasbusa:
https://github.com/sigp/lighthouse/issues/7424#issuecomment-2863228412


pk910 (approved):

pawanjay176:
This just made the default target-peers 0 and doesn't allow changing it

PR #1040, opened 2025-05-23
feat: add txpool-viz

ndeto:
This PR adds [txpool-viz](https://github.com/punkhazardlabs/txpool-viz) as an additional service to this package


barnabasbusa:
I know the PR is still in draft, but I think it would be good if txpool viz params would have ways to modify default docker image, add extra runtime params, and add env variables. (Similar to all our toolings)

ndeto:
> I know the PR is still in draft, but I think it would be good if txpool viz params would have ways to modify default docker image, add extra runtime params, and add env variables. (Similar to all our toolings)

Hi @barnabasbusa , This is ready for a review.

barnabasbusa:
Could you please add txpool_viz to 

```
	modified:   .github/tests/mix-public.yaml
	modified:   .github/tests/mix-with-tools-mev.yaml
	modified:   .github/tests/mix-with-tools-minimal.yaml
	modified:   .github/tests/mix-with-tools.yaml
```

And make sure the CI jobs are not failing

barnabasbusa:
Also it seems like you are missing default txpool viz parameters in input parser.

Probably want something like:
```python
def get_default_txpool_viz_params():
    return {
        "image": constants.DEFAULT_TXPOOL_VIZ_IMAGE,
        "min_cpu": False,
        "max_cpu": False,
        "min_mem": False,
        "max_mem": False,
        "extra_args": [],
        "polling": {
            "interval": "0.5s",
            "timeout": "3s",
        },
        "filters": {
            "min_gas_price": "1gwei",
        },
        "focil_enabled": "false",
        "log_level": "info",
        "env": {},
    }

```

and set the txpool_viz_params by default in Line 1067

ndeto:
@barnabasbusa , Kindly trigger the tests. I have also granted write access from the fork side incase you need to trigger tests again.

PR #1048, opened 2025-05-29
fix: pull kurtosis images in kurtosis install

tedim52:


barnabasbusa:
could you add all the images while at it? Logger vector images and whatever else we might need

barnabasbusa (approved):

PR #1086, opened 2025-06-27
feat: add node index label

0xTylerHolmes:
This PR adds a new label to the services that run for each node which is meant to be a quality of life improvement. The new label node-index which allows you to query all services for a node with a simple selector versus parsing the service-name.


src/cl/grandine/grandine_launcher.star
```diff
@@ -365,7 +365,8 @@ def get_beacon_config(
             client_type=constants.CLIENT_TYPES.cl,
             image=participant.cl_image[-constants.MAX_LABEL_LENGTH :],
             connected_client=el_context.client_name,
-            extra_labels=participant.cl_extra_labels,
+            extra_labels=participant.cl_extra_labels
```
barnabasbusa, line 368: could you just pass participant_index to the label maker and add it there?

0xTylerHolmes: yes this is my intent;. we are adding it to the extra_labels parameter which is used in the call to the label_maker. I didn't want to change the signature of the label_maker because not all services which use the label maker will have  a node-index.

0xTylerHolmes: Without changing the signature of the of the label_maker though the function won't know what the current participant index. My understanding is that the label_maker should be usable for any service, including non-indexed services, don't you think its cleaner to no change the signature? If you'd like I can make a new label_maker that includes a node-index arg. I do think it kind of abuses the "extra_labels" field as it didn't come from the configuration file. It just felt like the easiest place to add it.

barnabasbusa: why is it a big deal to change the signature of the label maker tho? :D

barnabasbusa (approved):

PR #1090, opened 2025-06-29
chore(docs): Teku documentation link for multi-domain and IP support in generate_certs.sh

VolodymyrBg:
Replaced the outdated Teku documentation URL with the current, working link for the section on supporting multiple domains and IPs in the key manager API. This ensures users are directed to the latest official instructions for generating certificates compatible with Teku's requirements.


barnabasbusa:
Sorry too low effort PR.

PR #1103, opened 2025-07-05
Fix: use correct port key for blobber validator proxy in launcher

VolodymyrBg:
Replaced the numeric port key (BLOBBER_VALIDATOR_PROXY_PORT_NUM) with the string identifier (BLOBBER_VALIDATOR_PROXY_PORT_ID) when accessing the validator proxy port in the blobber service context. This ensures the correct port value is retrieved from the service's ports dictionary, preventing potential runtime errors due to key mismatch.


barnabasbusa:
@parithosh this fix still needed? If not lets close the PR

PR #1106, opened 2025-07-08
feat: add log_level as cli for grandine

barnabasbusa:
Blocked on https://github.com/grandinetech/grandine/pull/249


barnabasbusa:
Guess this is not happening.

PR #1146, opened 2025-08-08
feat: besu: remove sync-mode config logic

macfarla:
Removing this setting from here enables us to override sync-mode via kurtosis config, if required for testing.

Also this exact same logic now exists in besu itself - 

default: SNAP if a --network is supplied. FULL otherwise.

### DIFF
--- a/src/el/besu/besu_launcher.star
+++ b/src/el/besu/besu_launcher.star
@@ -151,11 +151,6 @@ def get_config(
         "--engine-jwt-secret=" + constants.JWT_MOUNT_PATH_ON_CONTAINER,
         "--engine-host-allowlist=*",
         "--engine-rpc-port={0}".format(ENGINE_HTTP_RPC_PORT_NUM),
-        "{0}".format(
-            "--sync-mode=FULL"
-            if network_params.network in constants.NETWORK_NAME.kurtosis
-            else "--sync-mode=SNAP"
-        ),
         "--metrics-enabled=true",
         "--metrics-host=0.0.0.0",
         "--metrics-port={0}".format(METRICS_PORT_NUM),


macfarla:
to test this, here's a config passing sync-mode through to besu

```
participants_matrix:
  el:
    - el_type: besu
      el_image: ethpandaops/besu:fusaka-devnet-3
      el_extra_params: [--sync-mode="FULL"]
  cl:
    - cl_type: teku
      cl_image: ethpandaops/teku:fusaka-devnet-3
network_params:
  network: fusaka-devnet-3
persistent: true
```

without this PR, you get this error -
```
  == SERVICE 'el-1-besu-teku' LOGS ===================================
  Listening for transport dt_socket at address: 40043
  option '--sync-mode' (<MODE>) should be specified only once
```

macfarla:
I tried to test this with 
`kurtosis run --enclave fusakadevnet3 github.com/macfarla/ethereum-package/patch-1 --args-file ./fusaka-devnet-3-minimal-kurtosis.yaml`
but I get an error
```
There was an error interpreting Starlark code
No 'kurtosis.yml' file was found in the package root so fell back to Docker Compose package, but no default Compose files (compose.yml, compose.yaml, docker-compose.yml, docker-compose.yaml, docker_compose.yml, docker_compose.yaml) were found. Either add a 'kurtosis.yml' file to the package root or add one of the default Compose files.
```

macfarla:
some discussion w besu devs - 
"--sync-mode=FULL" if network_params.network in constants.NETWORK_NAME.kurtosis
^ what does that condition mean exactly? Any named network or some special kurtosis one?  

Thought you had to add this to get SNAP, before we made it default for named networks in Besu. So I was thinking this is a workaround that's no longer required - but do we need this to get SNAP on devnets?

being able to override the default config would be ideal - this came up because I wanted to do a FULL sync on a devnet and had to hack my local besu build to do it.

barnabasbusa (approved):

barnabasbusa:
is this good to be merged?

PR #1150, opened 2025-08-12
feat: add Tempo as an additional service to collect Lighthouse tracing data

jimmygchen:
This PR adds `tempo` as an additional service, and adds the collector url to lighthouse startup parameters, so that lighthouse nodes can push traces to Tempo tracing backend when it's enabled.

When tempo is enabled, Lighthouse traces can be observed in the Grafana `Explore` tab, and select `Search` query type:

image

I have a separate branch identical to this one but with the `kurtosis.yml` file modified to my fork, so it can be run with:
```
kurtosis run --enclave local-testnet github.com/jimmygchen/ethereum-package@tempo-test --args-file ./network_params.yaml
```

Note: this PR is coded by Claude and it's working - I'm using it but I likely won't have capacity in the short term to polish it further, feel free to leave the PR open or take over. 

Another current limitation is that the service name is hard coded to `lighthouse` so we can't distinguish between different lighthouse nodes. We could add a flag to lighthouse to configure the tracing service name.

### DIFF
--- a/src/cl/grandine/grandine_launcher.star
+++ b/src/cl/grandine/grandine_launcher.star
@@ -53,6 +53,7 @@ def launch(
     participant_index,
     network_params,
     extra_files_artifacts,
+    tempo_otlp_grpc_url=None,
 ):
     config = get_beacon_config(
         plan,
@@ -74,6 +75,7 @@ def launch(
         participant_index,
         network_params,
         extra_files_artifacts,
+        tempo_otlp_grpc_url,
     )
 
     beacon_service = plan.add_service(beacon_service_name, config)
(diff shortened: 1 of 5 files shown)


barnabasbusa:
you can use the service name instead of ip, within the same docker network dns will work.

jimmygchen:
I've changed it to use service name instead.

If this feature ends up being accepted, the next thing that would be useful is to configure metric generation for traces - that would give us metrics for timing / count of spans. I didn't attempt to do it in this PR because it would require changes to the prometheus package to enable remote write API, e.g.:

https://github.com/jimmygchen/lighthouse-metrics/blob/9f2af74c7f8daad8cacc8144216d8acb411b3ad4/docker-compose.yaml#L19

Here's a working example of the metrics generator tempo config, that generates metrics and push to tempo:
https://github.com/jimmygchen/lighthouse-metrics/blob/9f2af74c7f8daad8cacc8144216d8acb411b3ad4/tempo.yaml#L19-L40

`metrics_generator.processor.span_metrics.dimensions` would be a useful param to add to override params.

PR #1158, opened 2025-08-18
fix: corrected prysm supernode flag

CedArctic:
The previous flag was incorrect, and prevented prysm from starting. The correct flag is `--subscribe-all-subnets`. Source: [prysm flags documentation](https://prysm.offchainlabs.com/docs/configure-prysm/parameters/).

### DIFF
--- a/src/cl/prysm/prysm_launcher.star
+++ b/src/cl/prysm/prysm_launcher.star
@@ -211,7 +211,7 @@ def get_beacon_config(
     ]
 
     supernode_cmd = [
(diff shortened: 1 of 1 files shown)


barnabasbusa (changes requested):
subscribe all subnets and subscribe all data subnets is not the same thing

CedArctic:
You're right, it makes more sense as a cl flag. I am simulating setups where I want nodes with no vaildators to be subscribed to all subnets, so this flag comes in handy there.

PR #1187, opened 2025-09-19
feat: support amsterdam time

jihoonsong:
This PR supports Amsterdam time. Features for Glamsterdam such as [BALs](https://github.com/ethereum/go-ethereum/pull/32263) could use this.


barnabasbusa:
Amserdam Time is already in genesis.json and geth should be able to read it from that, assuming you set gloas_fork_epoch.

PR #1197, opened 2025-09-26
feat: Add Kurtosis config for BALs devnet 0

jihoonsong:


.github/tests/bal-devnet-0.yaml_norun
```diff
@@ -0,0 +1,38 @@
...
+    cl_type: prysm
+    cl_image: ethpandaops/prysm-beacon-chain:bal-devnet-0
+    vc_type: prysm
+    vc_image: ethpandaops/prysm-validator:bal-devnet-0
+    count: 2
+network_params:
+  genesis_delay: 20
+  fulu_fork_epoch: 0
+  gloas_fork_epoch: 1
+  perfect_peerdas_enabled: true
```
barnabasbusa, line 17: You gonna need a lot more nodes for this to work. 

Any reason you want this enabled?
You probably better off just setting your first participant to a supernode.

barnabasbusa: just enabled besu, and reth should be enabled already.

jihoonsong: Added both! Now it has clients of (Besu, Geth and Reth) x (Lodestar and Prysm).

PR #1208, opened 2025-10-02
feat: add `depends_on` to store service files for parallel

tedim52:
To support the use of the `--parallel` flag on Kurtosis v1.12.1, a `depends_on` to `store_service_files` to ensure that `store_service_files` runs after exec instructions.

This also adds `dns_name` to `el_context`. This allows distinguishing to use `dns_name` (service.name under the hood) or `ip_addr` (service.ip_address). Using context.ip_addr will create a dependency under the hood whereas `dns_name` won't

### DIFF
--- a/.github/tests/mev-pectra.yaml
+++ b/.github/tests/mev-pectra.yaml
@@ -55,6 +55,8 @@ network_params:
   shard_committee_period: 1
   churn_limit_quotient: 16
   prefunded_accounts: '{"0xb9e79D19f651a941757b35830232E7EFC77E1c79": {"balance": "100000ETH"}}'
+  electra_fork_epoch: 0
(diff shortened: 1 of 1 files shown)


barnabasbusa:
@tedim52 this good to be reviewed?

src/snooper/snooper_el_launcher.star
```diff
@@ -68,7 +68,7 @@ def launch_snooper(
 
     snooper_service = plan.add_service(snooper_service_name, snooper_config)
     return snooper_el_engine_context.new_snooper_el_client_context(
-        snooper_service.name,
+        snooper_service.ip_address,
```
barnabasbusa: did you wanna do this?

PR #1238, opened 2025-11-01
feat: add support for separate bootnode with bootnodoor

pk910:


barnabasbusa:
can you add some readme info too?

src/bootnodoor/bootnodoor_launcher.star
```diff
@@ -0,0 +1,188 @@
...
+        node_selectors=node_selectors,
+        tolerations=tolerations,
+    )
+
+
+def get_el_genesis_hash(plan, network_params, el_cl_genesis_data):
+    # For mainnet and sepolia, use static genesis hashes
+    if network_params.network == "mainnet":
+        return "0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3"
+    elif network_params.network == "sepolia":
```
barnabasbusa: any reason we don't need static genesis hashes for hoodi?

pk910: hoodi & all our devnets have a post-bellatrix genesis,  and therefore the genesis block hash is included in deposit_contract_block_hash.txt
for mainnet & sepolia we're not storing the genesis block hash in the repository

PR #1274, opened 2025-12-23
fix: missing dns_name

parithosh:
The `main` branch was failing when `snooper_enabled: false` was set. This was due to the missing `dns_name` field in snooper context structs that was overlooked in #1208.

  Changes:
  - Added dns_name field to snooper_el_context.star and snooper_beacon_context.star
  - Updated snooper_el_launcher.star to pass snooper_service.name as dns_name
  - Updated snooper_beacon_launcher.star to pass snooper_service.name as dns_name (also fixed: was passing .name instead of .ip_address for ip_addr)


barnabasbusa (approved):
not sure how i missed that

PR #1277, opened 2026-01-01
fix: support custom CL for MEV builder (issue #627)

### LINKED ISSUE #627: Cannot configure non-Lighthouse CL for builder
This isn't really an issue, but I noticed that we are forced to use Lighthouse as the CL for the builder. It could be useful to test with other CLs too.

DarkLord017:
- Infer CL type from mev_builder_cl_image instead of hardcoding Lighthouse
- Set CL-specific builder flags for all supported clients
- Use correct beacon HTTP port in builder config CLEndpoint
- Add robust cl_type inference in cl_launcher based on image

Tried to fix #627 
Please review @parithosh @barnabasbusa , am i heading towards the issue correctly

### DIFF
--- a/src/cl/cl_launcher.star
+++ b/src/cl/cl_launcher.star
@@ -122,34 +122,8 @@ def launch(
     cl_service_configs = {}
     cl_participant_info = {}
     for index, participant in enumerate(args_with_right_defaults.participants):
-        # The declared client type from the participant YAML
-        declared_cl = participant.cl_type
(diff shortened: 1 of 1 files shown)


DarkLord017:
> I still do not see how will nimbus and grandine going to trigger a payload event for each slot?

I looked at their docs & code and tried to find out , but wasn't able to .. is it possible that it does not exst for these clients ?

PR #1352, opened 2026-04-06
feat: Have a dedicated node for buildoor

bharath-123:


pk910:
As discussed in mattermost:
buildoor currently uses the first regular participant as backend. It uses the standard apis (el rpc, el engine & cl rpc), with no custom image need. It doesn't cause any negative side effects on the used node either.
Under these conditions, I'd prefer keeping the existing logic to just pick the first participant as buildoor backend.
It feels more correct to keep the client configuration in one place and it's easier to swap around and test different client pairs.

barnabasbusa:
I agree with pk, I'd prefer not to do this.

bharath-123:
My reasoning to have a dedicated node for buildoor is that the participants in kurtosis are usually validator running nodes which is usually not a setup real builders would run with. A real builder would run their block building software next to a full node. So giving buildoor a fully dedicated node made sense to me as it is how we would run it on a devnet..
I think we could explicitly not have the first node to be a validating node in this setup if required. So I am fine with either approach for now.

barnabasbusa:
You can explicitly disable validators on your first node with validator_count: 0 if you want. So if this is the only reasoning to have a dedicated node for it, then its not a very strong one imo.

bharath-123:
I didn't know about `validator_count`. If we can do that, then i will close this PR. Thanks!

PR #1353, opened 2026-04-07
feat: support gpu ere prover in zkboost

frisitano:
# Overview
This PR introduces ere gpu prover support and gpu count validation when we have multiple gpu services.

depends on kurtosis version [1.18.1](https://github.com/kurtosis-tech/kurtosis/releases/tag/1.18.1)


barnabasbusa:
Can you add 2 examples in 

.github/tests/1gpu_zkvm.yaml_norun
and an 8gpu example too? 

norun as we don't want them to ever run on a ci, but would be good to be able to point to a specific file in case someone wants to get up and running with an example with 1/8 gpus

barnabasbusa (approved):

PR #1360, opened 2026-04-17
fix(stability-check): set maxForkDistance to 3 to tolerate transient head disagreements

yperbasis:
## Summary

The bundled `static_files/assertoor-config/tests/stability-check.yaml`
defines `check_consensus_forks` with no `config:` block. As a result, it
falls through to assertoor's `DefaultConfig()`:

```go
// ethpandaops/assertoor pkg/tasks/check_consensus_forks/config.go
func DefaultConfig() Config {
    return Config{
        MinCheckEpochCount: 1,
        MaxForkDistance:    1,
    }
}
```

`MaxForkCount` is therefore the Go zero value, **0**, which combined
with `MaxForkDistance: 1` means that any transient head-view
disagreement between two CL clients within a single slot distance trips
the check.

In multi-CL-type devnets (e.g. teku + lighthouse) with `seconds_per_slot`
in the 6–8s range, such momentary disagreements are routine and nearly
unavoidable on a small validator set. This is the primary flake signature
we observe in [Erigon's kurtosis-pectra CI](https://github.com/erigontech/erigon/actions/runs/24561747013/job/71811815555):

```
check failed: too many forks. (have: 1, want <= 0)
```

## Change

Add an explicit `config:` block under `check_consensus_forks` setting
`maxForkDistance: 3`. This tolerates transient head lag up to 3 slots
while leaving `MaxForkCount` at its default of 0, so any genuinely
divergent fork still fails fast.

Note: the pass/fail in assertoor is `len(headForks) - 1 > MaxForkCount`,
so a non-zero `maxForkCount` would let up to N simultaneous divergent
heads slip by regardless of their length. Keeping it at 0 is the
stricter and safer choice.

## Test plan

- [x] `git diff` shows only the 2-line config addition
- [ ] Sanity-check from a maintainer on whether 3-slot tolerance is the
      right default


pk910 (approved):

yperbasis:
Added a follow-up commit d7d4aa4 dropping `maxForkCount: 3`, keeping only `maxForkDistance: 3`.

Rationale: the pass/fail check in assertoor is `len(headForks) - 1 > MaxForkCount`, so `maxForkCount: 3` tolerates up to 3 simultaneous divergent heads regardless of their length — including forks longer than `maxForkDistance` that the test should actually catch. Leaving `MaxForkCount` at its default of 0 fails fast on any genuinely divergent head, while transient lag is still absorbed by the canonical-block grouping inside `GetHeadForks`.

Let me know if you'd prefer I squash before merge — happy to do either.

@pk910 sorry for the flip-flop; this landed before I reviewed the assertoor task code carefully.

pk910 (approved):

PR #1376, opened 2026-04-28
feat(prysm): default --p2p-max-peers to participant count

barnabasbusa:
## Summary
- Set Prysm beacon's `--p2p-max-peers` to the total number of CL participants so the peer cap scales with the kurtosis network size without manual `cl_extra_params` tuning.
- Plumb `num_participants` through `new_prysm_launcher` so it's available inside `get_beacon_config`.

## Test plan
- [ ] Spin up a multi-participant network including a Prysm node and verify `--p2p-max-peers=<num_participants>` appears in its launch cmd.
- [ ] Confirm Prysm starts cleanly and peers up across a small (e.g. 2-3) and larger (e.g. 8+) participant set.
- [ ] Verify other CL clients (lighthouse/teku/nimbus/lodestar/grandine) are unaffected.

### DIFF
--- a/src/cl/cl_launcher.star
+++ b/src/cl/cl_launcher.star
@@ -72,6 +72,7 @@ def launch(
             "launcher": prysm.new_prysm_launcher(
                 el_cl_data,
                 jwt_file,
+                num_participants,
             ),
             "launch_method": prysm.launch,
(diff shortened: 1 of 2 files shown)


skylenet:
no

PR #1385, opened 2026-05-05
feat(zkboost): in-process verifier kind + per-instance zkvms

qu0b:
## Summary

Adds two improvements to the kurtosis zkboost integration:

1. **`kind: verifier`**: zkboost can now run as an in-process verifier without any backing `ere-server`. Mirrors the new variant in eth-act/zkboost#61 (depends on that PR).
2. **Per-instance `zkvms`**: each entry in `zkboost_params.instances` may define its own `zkvms` list, falling back to the global one if not set. This is needed to mix prover and verifier zkboost instances in the same kurtosis network.

## Why

`ere-server` is heavy: it loads the full ZisK prover circuit into 96 GiB VRAM on init, holds 54 GiB of host RAM, and ships a 24.4 GB Docker image. EIP-8025 only needs a node to *verify* gossiped proofs, which is a CPU-only ~5 ms operation against a 32-byte verifying key. With this change, kurtosis can stand up:

- prover participants → `kind: ere` → real ere-server + GPU
- verifier-only participants → `kind: verifier` → 11 MiB zkboost process, **no GPU, no ere-server**

This unlocks single-GPU multi-prover-network testing, makes verifier-only nodes runnable on contributor laptops without GPUs, and matches the per-node capability model the EIP-8025 spec describes (`proof_types` is a dynamic per-node advertisement, not a global protocol constant — see `consensus-specs/specs/_features/eip8025/p2p-interface.md`, `ExecutionProofStatus`).

## Changes

- `src/zkboost/zkboost_launcher.star`:
  - Auto-resolve `program_vk_url` from zkboost's pinned ere-guests version (new `ERE_GUESTS_VK_URL_TEMPLATE`).
  - Per-instance zkvms: `instance.get(\"zkvms\", zkboost_params.zkvms)`. ere-server launches collected across all instances, deduped by `proof_type`. Verifier entries skip ere-server entirely.
  - Render new `kind: verifier` template entries with `program_vk_url`.
- `src/package_io/input_parser.star`: add `verifier` to valid kinds.
- `static_files/zkboost-config/config.toml.tmpl`: render `kind = \"verifier\"` with `program_vk_url`.
- `.github/tests/zkboost_in_process_verifier.yaml_norun` (new): demo config — one ere-prover participant + one verifier-only participant, both for `reth-zisk`.

## Test plan

End-to-end on a kurtosis enclave on an RTX PRO 6000 Blackwell rig (1 prover + 1 verifier, 30 s slots, ~14 hours of blocks):

| Side | Container RAM | GPU VRAM | Per-op time | Image |
|---|---|---|---|---|
| prover (`kind: ere`) | 54 GiB (ere-server) + 60 MiB (zkboost) | 96 GiB | 6.37 s/proof avg | 24.4 GB + 180 MB |
| verifier (`kind: verifier`) | 11 MiB | **0** | **5.6 ms/verify avg** | 180 MB |

Verifier metrics: `zkboost_verify_total{proof_type=\"reth-zisk\",verified=\"true\"} 3252`, `zkboost_verify_duration_seconds_sum / count = 5.6 ms`. Same block hashes verified across both sides.

## Dependency

Depends on **eth-act/zkboost#61**. Until that lands, the test config references a custom image (`zkboost:cpu-verify-experiment`) built from the linked branch. Once upstream zkboost ships the new variant in an official image, the test config switches to `ghcr.io/eth-act/zkboost/zkboost:<version>` and auto-resolution (already supports `program_vk_url`) handles the rest.

Marking this PR as draft until the zkboost PR merges.

## Follow-ups

- Upstream zkboost: feature-gate verifier crates per zkVM (currently only `verifier-zisk`).
- CI assertion shape: probe `zkboost-verifier:3000/metrics` after N slots, expect `zkboost_verify_total > 0`.
- Lighthouse-side: BN-only gossip subscription flag so verifier-only deployments don't need `vc_extra_params --proof-types` (which currently makes the VC hit zkboost's prove path and get rejected — visible as `zkboost_prove_total{status=\"error\"}` on the verifier).
- Per-instance zkvm validation: `_validate_ere_gpu_config` and the zkvm validator in `input_parser.star` only check the global list. Per-instance zkvms aren't validated upfront. Acceptable for the prototype; should be tightened.


han0110:
The in-process verifier is released in https://github.com/eth-act/zkboost/releases/tag/v0.7.0, and tested locally the newly added `zkboost_in_process_verifier.yaml_norun‎` works with the latest image tag.

PR #1393, opened 2026-05-15
feat(otel): in-enclave ClickHouse + log bridge for OTel-shaped log capture

samcm:
Adds an `otel` additional_service: single-node ClickHouse, a Go bridge that streams logs from the Kurtosis engine's `GetServiceLogs` RPC into `otel.otel_logs`, and an OpenTelemetry Collector accepting OTLP traces into `otel.otel_traces`. Every EL/CL/VC participant is automatically configured with the standard `OTEL_EXPORTER_OTLP_*` env vars when the service is enabled, so OTel-SDK-aware clients ship traces with no extra config. Schema mirrors the platform `clickhouse-observability` shape minus the IngressUser / k8s materialized columns. When `grafana` is also enabled, the in-enclave ClickHouse is auto-registered as a Grafana datasource via the `grafana-clickhouse-datasource` plugin.


barnabasbusa:
would you mind adding geth's otel while you are at it?

samcm:
Superseded by #1405, which moves the OTel collector + ClickHouse to the Kurtosis engine level (kurtosis-tech/kurtosis#3122) instead of running them in-enclave, and replaces this in-enclave bridge with opt-in per-client tracing. Closing in favour of that approach.

PR #1394, opened 2026-05-15
feat(zkboost): set default proof_timeout_secs to seconds_per_slot

han0110:
If longer `network_params.seconds_per_slot` is configured, it might make sense for the prover to have more time to create the proof by default.


barnabasbusa:
Would you actually have seconds per slot time or would you need to take a percentage of this? I thought currently with 12s/slot we could have like 9s to create a proof?

han0110:
Yeah 9s proving time in 12s slot context is more practical.

But I assume when needing to set slot time >12s is because the configured prover can't create the proof timely, I think this PR is mainly to remove the issue that when we set longer slot time without setting `proof_timeout_secs` in zkboost param, we still see proof timeout (which is a bit annoying).

But on the 9s you mentioned, I think we can also make the default `proof_timeout_secs` to be 3/4 of slot time, if slot time <=12s (not sure if this fixed percentage makes sense tho).

barnabasbusa:
yeah let's do 3/4 x seconds per slot

han0110:
Updated in https://github.com/ethpandaops/ethereum-package/pull/1394/changes/37f2c0443db78230657523bcadec7d42e1f514e8

barnabasbusa (approved):
Verified locally: with `network_params.seconds_per_slot: 13` and zkboost enabled, the rendered `/config/config.toml` on the zkboost container shows `proof_timeout_secs = 9` (13 * 3 // 4), confirming Starlark evaluates the expression correctly.

PR #1401, opened 2026-05-19
fix(zkboost): fill mock_proving_time defaults from slot duration

barnabasbusa:
## Summary

- A partial override like `mock_proving_time: { kind: constant }` silently became 0ms because dict.get on the missing `ms` key returned 0, while omitting `mock_proving_time` entirely fell back to the launcher's hardcoded 6000ms — surprising and inconsistent.
- Fill in kind-appropriate defaults during input validation so partial overrides behave like the no-override case. Defaults now scale with `slot_duration_ms` (2/3 of slot): `constant.ms`, `random.min_ms`/`max_ms` (default_ms/2 .. default_ms*2), and `linear.ms_per_mgas`.
- Updated the launcher's defensive fallback and the auto-injected default zkvm to use the same 2/3-of-slot scaling.

## Test plan

- [ ] `kurtosis lint --format .` passes
- [ ] `zkvms: [{ kind: mock, proof_type: reth-zisk }]` → mock proving time defaults to 4000ms on 6s slots (2/3 of 6000)
- [ ] `zkvms: [{ kind: mock, proof_type: reth-zisk, mock_proving_time: { kind: constant } }]` → resolves to 4000ms (no longer 0ms)
- [ ] `zkvms: [{ kind: mock, proof_type: reth-zisk, mock_proving_time: { kind: random } }]` → resolves to 2000–8000ms on 6s slots
- [ ] On a 12s slot, defaults scale to 8000ms / 4000–16000ms

### DIFF
--- a/src/zkboost/zkboost_launcher.star
+++ b/src/zkboost/zkboost_launcher.star
@@ -138,7 +138,11 @@ def launch_zkboost(
                 entry["ProgramVkUrl"] = zkvm["program_vk_url"]
             elif zkvm["kind"] == "mock":
                 mock_proving_time = zkvm.get(
-                    "mock_proving_time", {"kind": "constant", "ms": 6000}
+                    "mock_proving_time",
+                    {
+                        "kind": "constant",
+                        "ms": network_params.slot_duration_ms * 2 // 3,
+                    },
                 )
                 entry["MockProvingTimeKind"] = mock_proving_time.get("kind", "constant")
                 entry["MockProvingTimeConstantMs"] = mock_proving_time.get("ms", 6000)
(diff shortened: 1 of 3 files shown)


src/zkboost/zkboost_launcher.star
```diff
@@ -138,7 +138,11 @@ def launch_zkboost(
...
                 mock_proving_time = zkvm.get(
-                    "mock_proving_time", {"kind": "constant", "ms": 6000}
+                    "mock_proving_time",
+                    {
+                        "kind": "constant",
+                        "ms": network_params.slot_duration_ms * 2 // 3,
+                    },
                 )
                 entry["MockProvingTimeKind"] = mock_proving_time.get("kind", "constant")
                 entry["MockProvingTimeConstantMs"] = mock_proving_time.get("ms", 0)
```
chetanyb: Verified locally.

Minor suggestion: Since input_parser now guarantees all fields are filled, the launcher fallbacks are now dead code.

chetanyb (changes requested):

chetanyb (approved):

PR #1414, opened 2026-06-07
feat: enable genesis validator shuffling & add proper name mapping

pk910:
gated on https://github.com/ethpandaops/ethereum-genesis-generator/pull/296

Enables genesis validator shuffling by default and maps the shuffled on-chain indices back to client names.

- New `shuffle_genesis_validators` network param (default `true`), passed through to the genesis generator.
- The validator-ranges generator now reads `validator_names.yaml` from the genesis artifact **at execution time** and translates the mnemonic-derivation index ranges into on-chain index ranges per participant — splitting where client boundaries don't align with the mapping groups, and passing non-main-mnemonic sources through under their own name.
- Backward compatible to an EGG without a mapping file: it falls back to the direct 1:1 index mapping.
- Removes the now-unused validator-ranges Go template (ranges are written directly).


barnabasbusa:
pls bump egg to 4.0.8

barnabasbusa (approved):

PR #1421, opened 2026-06-14
fix: set --http-mev-relay url to buildoor url

bharath-123:

PR #1422, opened 2026-06-14
fix: fix buildoor wiring (skip mev-boost)

pk910:
replacement for #1421 with a proper fix to skip mev-boost entirely instead of just putting a patch on the prysm flag :)


bharath-123 (approved):
this is indeed much better! lets get it in!

barnabasbusa (approved):

PR #1455, opened 2026-07-24
feat: state_actor_params — pre-populate EL databases with state-actor

qu0b:
## Summary

Adds an opt-in `state_actor_params` package arg that pre-populates EL client databases with [state-actor](https://github.com/ethereum/state-actor) before the EL clients launch.

When enabled, one state-actor generation run happens per unique `el_type` among the participants — after genesis generation, driven by the network's own `genesis.json` (`state-actor --genesis`) — so the produced database's genesis hash matches the hash the CL genesis was built against, and no CL-side changes are needed. Each run's output datadir is stored as a files artifact named `state-actor-<el_type>-data` and registered so participants can opt in by mounting it at their execution data dir:

```yaml
participants:
  - el_type: geth
    cl_type: lighthouse
    el_extra_mounts:
      /data/geth/execution-data: state-actor-geth-data
state_actor_params:
  enabled: true
```

This follows the shadowfork pattern (pre-launch step fills the EL data location before the client starts), but generates state instead of downloading a snapshot.

## Why

state-actor writes client-native databases (Pebble for geth, MDBX/RocksDB for reth, Bonsai RocksDB for besu, flat-state for nethermind, RocksDB for ethrex, Erigon v3 snapshots) without going through each client's init path — useful for state-bloat experiments and cross-client determinism testing. This PR wires its output into participant startup. Alloc-verbatim mode (no `target_size`) is fully supported today; bloated-state mode additionally requires anchoring the CL genesis to the state-actor genesis hash, which is left for a follow-up (state-actor exports an `eth_getBlockByNumber`-shaped `genesis_block.json` for exactly that purpose).

## Changes

- `src/state_actor/state_actor_launcher.star` (new): runs the per-el_type generation via `plan.run_sh`, stores + registers the datadir artifacts
- `src/participant_network.star`: invoke the launcher after `el_cl_data` is generated, before EL launch
- `src/package_io/input_parser.star`, `src/package_io/sanity_check.star`: `state_actor_params` plumbing (`enabled`, `seed`, `target_size`, `images`)
- `README.md`: config reference entry

## Testing

Tested with one single-node enclave per supported EL (geth, reth, besu, nethermind, ethrex, erigon), each paired with lighthouse, using state-actor images built from [ethereum/state-actor compare: qu0b/ethereum-package-integration](https://github.com/ethereum/state-actor/compare/main...qu0b:state-actor:qu0b/ethereum-package-integration) (adds the `--genesis` flag; not yet upstream — until it merges, `state_actor_params.images` must point at images built from that branch):

- all six ELs booted directly from the pre-populated database (verified via manifest/DB files in the datadir and rotated RocksDB logs) with no init step
- every client's RPC `eth_getBlockByNumber(0x0)` hash equals the hash state-actor exported, so the existing CL genesis stayed valid
- all six clients produced the identical genesis state root for the same 283-account alloc
- all networks produced blocks with lighthouse (heads observed: geth 20, reth 12, besu 17, nethermind 14, ethrex 15, erigon 13) with no EL errors

Per-client notes surfaced by testing (documented in the README entry): nethermind participants need `--Init.BaseDbPath=<datadir>` + `--FlatDb.Enabled=true`; erigon's pre-launch `erigon init` coexists cleanly with the pre-seeded datadir.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01TJoWPhxuJVSwvzpZr494X1

### DIFF
--- a/README.md
+++ b/README.md
@@ -1474,6 +1474,46 @@ docker_cache_params:
   google_prefix: "/gcr/"
 
 
+# Pre-populate EL client databases with state-actor (https://github.com/ethereum/state-actor)
+# before the EL clients launch. When enabled, one state-actor run happens per unique
+# el_type among the participants, driven by the network's own genesis.json
+# (state-actor --genesis), so the produced database's genesis hash matches the hash the
+# CL genesis was built against. Each run's output is stored as a files artifact named
+# "state-actor-<el_type>-data"; participants opt in by mounting it at their execution
+# data directory via el_extra_mounts:
+#
+#   participants:
+#     - el_type: geth
+#       el_extra_mounts:
+#         /data/geth/execution-data: state-actor-geth-data
+#
+# nethermind participants additionally need:
+#   el_extra_params:
+#     - "--Init.BaseDbPath=/data/nethermind/execution-data"
+#     - "--FlatDb.Enabled=true"
+#
+# Supported el_types: geth, reth, besu, nethermind, ethrex, erigon
+# Requires state-actor images with --genesis support.
+state_actor_params:
+  # Whether to run state-actor pre-population
+  enabled: false
+  # Random seed passed to state-actor (deterministic output for a given genesis + seed)
+  seed: 1
+  # Optional DB-size budget (e.g. "1GB") for synthetic state on top of the genesis alloc.
+  # NOTE: any non-empty value changes the EL state root and therefore the genesis hash;
+  # the CL genesis must then be anchored to the state-actor output, which this
+  # integration does not do yet. Leave empty for alloc-verbatim databases.
+  target_size: ""
+  # Per-client state-actor images
+  images:
+    geth: ghcr.io/ethereum/state-actor-geth:main
+    reth: ghcr.io/ethereum/state-actor-reth:main
+    besu: ghcr.io/ethereum/state-actor-besu:main
+    nethermind: ghcr.io/ethereum/state-actor-nethermind:main
+    ethrex: ghcr.io/ethereum/state-actor-ethrex:main
+    erigon: ghcr.io/ethereum/state-actor-erigon:main
+
+
 # Configuration place for mempool bridge (https://github.com/ethpandaops/mempool-bridge)
 mempool_bridge_params:
   # The image to use for mempool bridge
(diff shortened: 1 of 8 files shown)


qu0b:
Rebased onto current main (#1447 refactor). Two adjustments along the way, both re-verified with a geth+lighthouse enclave on the rebased branch (EL boots from the pre-populated DB, RPC genesis hash matches state-actor's export, chain advances):

- the refactor removed `merge_nested_defaults`, so the `state_actor_params` parse branch now uses the same inline per-key merge as the neighbouring params (with per-client merging for `images`)
- the current ethereum-genesis-generator output omits a fork's `blobSchedule` entry when a BPO fork activates at the same timestamp (osaka+bpo1 at genesis); state-actor's genesis validation was stricter than current geth and rejected it — fixed on the state-actor branch ([qu0b/ethereum-package-integration](https://github.com/ethereum/state-actor/compare/main...qu0b:state-actor:qu0b/ethereum-package-integration)), so images must be rebuilt from that branch's tip.

qu0b:
Addressed the review feedback and extended the integration (commit 36f7886):

- **Full state-actor config support**: `state_actor_params.spec` takes an inline YAML state-actor spec (`--spec` — ERC-20s, EOAs, EIP-7702 accounts, raw bytecode; schema in state-actor's docs/SPEC.md), and `extra_args` passes arbitrary CLI flags through.
- **Bloated state now works end-to-end**: with `target_size` / `spec` set, the CL genesis is regenerated anchored to state-actor's genesis block (via the generator's existing `SHADOW_FORK_FILE` input, keeping the network's own network_id), so you can launch a chain with e.g. 10GB of state at block 0 and it finalizes normally. Participants set `el_pre_populated_db: true`, which makes launchers skip genesis init/override (geth `--override.genesis`, erigon `init`).
- **Snapshot reuse**: `state_actor_params.snapshots` takes per-client s3/https datadir tarball URLs and fetches instead of regenerating, through the same artifact/anchoring code paths. Requires pinning `network_params.genesis_time` (and identical network params) so the regenerated genesis matches the snapshot's.
- **persistent: true** writes datadirs straight into participants' persistent volumes (shadowfork-style helper services) — files artifacts can't carry multi-GB datadirs.

Verified on this branch: a geth+lighthouse network with 2GB synthetic state + a spec (bloated genesis accepted by the CL, spec EOA/ERC-20 queryable over RPC, blocks produced), and snapshot reuse over HTTP reproducing the source network's genesis hash byte-for-byte.

PR #1460, opened 2026-07-28
feat: Add standalone discv5 'bootnode' service

cskiraly:
Adds a dedicated discv5 bootnode enabled via additional_services: [bootnode]. It launches one EL node (geth by default, selectable via bootnode_params.el_type) whose sole job is discv5 discovery; with --maxpeers=0 it holds no RLPx peers and is a pure rendezvous. Its enode/ENR are fed into the existing bootnodoor override plumbing, so a single node seeds BOTH the EL and CL discovery overlays (discv5 is application-agnostic). A global bootnode_params.record switch selects the form (enr|enode) handed to EL clients.

Files:
- src/bootnode/bootnode_launcher.star (new): launches the bootnode, reuses geth.get_config
- src/participant_network.star: launch when enabled, feed EL+CL overrides
- src/package_io/input_parser.star: bootnode_params (el_type, el_image, el_extra_params, record)
- src/package_io/sanity_check.star: allow 'bootnode' service + bootnode_params
- src/el/el_launcher.star: global bootnode_record (enr|enode) selection

### DIFF
--- a/src/bootnode/bootnode_launcher.star
+++ b/src/bootnode/bootnode_launcher.star
@@ -11,6 +11,70 @@ shared_utils = import_module("../shared_utils/shared_utils.star")
 
 SERVICE_NAME = "bootnode"
 
+# Fixed identity for the devp2p bootnode (deterministic ENR across runs; devnet-only).
+DEVP2P_NODEKEY = "0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"
+DEVP2P_DISCOVERY_PORT = 30303
+
+
+# Launch a pure discv5 bootnode using go-ethereum's `devp2p discv5 listen` (no chain,
+# RLPx, RPC or txpool — just discv5). Needs an alltools image (e.g.
+# ethereum/client-go:alltools-latest) set as bootnode_params.image. It has no RPC to
+# query, so its ENR/enode are COMPUTED in-container with `devp2p key to-enr/to-enode`
+# from the fixed nodekey + the container's own IP. Returns (enode, enr).
+def _launch_devp2p_bootnode(plan, bootnode_params, node_selectors, tolerations):
+    config = ServiceConfig(
+        image=bootnode_params.image,
+        ports={
+            constants.UDP_DISCOVERY_PORT_ID: shared_utils.new_port_spec(
+                DEVP2P_DISCOVERY_PORT,
+                shared_utils.UDP_PROTOCOL,
+            ),
+        },
+        entrypoint=["devp2p"],
+        cmd=[
+            "discv5",
+            "listen",
+            "--nodekey",
+            DEVP2P_NODEKEY,
+            # Explicit empty bootnodes: suppress devp2p's DEFAULT (public mainnet)
+            # bootnodes, so the table isn't polluted with public nodes.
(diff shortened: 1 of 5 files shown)


MysticRyuujin:
Wouldn't the proper way to do this be to run an instance of `devp2p discv5 listen` instead of a full geth node?

https://github.com/ethereum/go-ethereum/tree/master/cmd/devp2p

Edit: Nevermind, I see it's configurable via `bootnode_params.el_type` so the scope is totally different

cskiraly:
> Wouldn't the proper way to do this be to run an instance of `devp2p discv5 listen` instead of a full geth node?
> 
> https://github.com/ethereum/go-ethereum/tree/master/cmd/devp2p
> 
> Edit: Nevermind, I see it's configurable via `bootnode_params.el_type` so the scope is totally different

Indeed, I'm adding the devp2p option now, after having been reminded by you and also by @MariusVanDerWijden that that should work as well. We could also make that the default, but I think keeping the other implementations there as options is a good way to check issues.

barnabasbusa:
We now default to using discv5 on every client. Do we still need this PR?

cskiraly:
This is not about using discv4 or discv5, but about configuring what implementation is used as bootnode.
- The package default is to mesh up all peers  (pass all nodes with a smaller ID)
- The only bootnode option is Bootnodoor, which is one specific implementation with tricks that are not there in a normal node
- This PR adds the possibility to use a discv5 compliant tool (devp2p tool from geth) or an EL client as bootnode.

pk910:
> The only bootnode option is Bootnodoor

I don't get this argument.  Bootnodoor is the only separately enableable bootnode that lives outside a client, 
but if not enabled the package uses the first participant as bootnode for all other participants.
So when configuring 2 pairs, eg.  lighthouse-geth + teku-besu ,  without enabling bootnodoor, the package will use lighthouse as CL bootnode and geth as EL bootnode.  Not via separate bootnode processes, but as part of the client stack (which shouldn't differ from the separate bootnode sub-commands?)

So, what's the difference between running the geth bootnode as standalone tool vs running geth as first participant so it's used as bootnode?

barnabasbusa:
Also ethereum package is frequently used for cutting edge testing like focil and frames. How will you make sure that the bootnode service that you are maintaining will be compatible with those features? (while those might not touch networking stack, there might be some other fork feature that someone would like to test which will require networking change). Would ideally like to reduce the number of dependences on outside tools needing to be updated for ethereum package to function.

cskiraly:
> but if not enabled the package uses the first participant as bootnode for all other participants.

Are you sure? I recall it is directly connecting the nodes, passing each node the address of every previous one.

cskiraly:
> Also ethereum package is frequently used for cutting edge testing like focil and frames. How will you make sure that the bootnode service that you are maintaining will be compatible with those features?

Clients tend to implement cutting edge features ;-) 

> there might be some other fork feature that someone would like to test which will require networking change
Yeah, clients also tend to be the first to prototype those networking changes.

The fact is there are many implementations of discv5, and the protocol is designed so that any of these could serve as a bootnode. It seems a good idea to me to support this variety, e.g. when checking protocol compatibility (while keep the default as is).

PR #1470, opened 2026-08-10
fix: derive builder keys past the preregistered validator range

shane-moore:
## Summary

Genesis ePBS builders derive their keys from the network mnemonic starting at the participants' validator sum (`actual_num_validators`). That is not the end of the validator keyspace: `preregistered_validator_count` can be larger, and those extra deposited-but-idle validators own the indices just past it. Builders then derive keys that are already in the genesis validator set, and `eth-beacon-genesis` aborts:

```
processing mnemonic 0, for 42 validators
processing mnemonic 1, for 2 validators
duplicate public key in validator set: 0xab40dc1c...77e5f2d at index 42
```

So `builder_count` currently cannot be combined with a `preregistered_validator_count` larger than the participants' validator sum. That surplus looks intentional rather than a misuse: the README documents the count as an override of the participant-derived value, and keystores are only generated over the participants' own ranges, so the extra validators are genesis entries that nothing in the package runs — the layout you get when something outside the package holds those keys.

- Builder keys now start at `max(actual_num_validators, preregistered_validator_count)`, resolved once in `parse_network_params` and carried as `network_params.builder_key_start_index`.
- `main.star` consumes that value for the `keys derived at indices X..Y` log line and for `instance_builder_key_index`, instead of recomputing from `total_validator_count`. The three sites can no longer drift apart.
- Resolved unconditionally rather than inside the `builder_count > 0` branch, since buildoor instances derive keys even when `builder_count` is `0`.

No behaviour change unless `preregistered_validator_count` is set larger than the participants' validator sum: it defaults to `0` and is only resolved from the participants later in `network_launcher/kurtosis.star`, so `max()` returns the previous start index for every other config.

## Testing

One participant, `validator_count: 32`, `builder_count: 2`, geth + lighthouse `glamsterdam-devnet-6`.

| `preregistered_validator_count` | branch | injected builder entry | genesis |
|---|---|---|---|
| 42 | main (`1bd4065`) | `"start": 32` | aborts: `duplicate public key in validator set: 0xab40dc1c... at index 42` |
| 42 | this branch | `"start": 42` | generated, proceeds to CL launch |
| unset | main (`1bd4065`) | `"start": 32` | generated |
| unset | this branch | `"start": 32` | generated |

No `.github/tests` config included, because I could not write one that passes. `builder_count` requires `gloas_fork_epoch: 0`, and I could not get a Gloas-at-genesis network past CL startup on any image I tried: `lighthouse:glamsterdam-devnet-6` fails with `Failed to build beacon chain: Head block not found in store`, and `glamsterdam-devnet-5` with `Built-in genesis state SSZ bytes are invalid`. That reproduces on `main` with no `builder_count` and no preregistered count at all, so it is independent of this change. It may also be why this path has no coverage today: no config under `.github/tests` sets `gloas_fork_epoch: 0`, and `builder_count` refuses to run at any other epoch. Happy to add a regression config if there is an image combination where Gloas-at-genesis comes up.

## Possible follow-up

Builders sharing the validator keyspace means every consumer has to know where the validator range ends, which is what went wrong here. Giving builders their own mnemonic would make that unrepresentable and drop the index arithmetic; `buildoor_launcher.star` already accepts `--builder-mnemonic`. Worth considering while `builder_count` and `buildoor_params.instances` are still unreleased, since changing the derivation later would move existing builder keys. Happy to follow up if you prefer that shape.


main.star
```diff
@@ -871,7 +873,7 @@ def run(plan, args={}):
             # not collide. The builder is onboarded after genesis via its lifecycle
             # deposit (buildoor_params.lifecycle), not registered at genesis.
             instance_builder_key_index = (
-                total_validator_count
+                network_params.builder_key_start_index
```
iurii-ssv, line 876: The per-participant buildoor path is fixed here, but the sibling site was missed: the deprecated single-shared `mev_type: buildoor` path at [`main.star:700`](https://github.com/shane-moore/ethereum-package/blob/c3b2cbfadd661d84b3e4bc9597f4c863f4972d17/main.star#L700) still passes `total_validator_count` as the buildoor's `--builder-key-index`.

Same trigger as the bug this PR fixes: with `preregistered_validator_count > actual_num_validators`, that buildoor derives its BLS key at `actual_num_validators` — inside the preregistered range — colliding with a preregistered genesis validator. It surfaces post-genesis (the lifecycle deposit lands on an already-registered key) rather than as a genesis abort, but it's the same root cause. After this PR, `total_validator_count` survives in exactly that one place.

Fix is the same swap there: `total_validator_count` → `network_params.builder_key_start_index`. Minimal on purpose — that path deliberately omits `+ builder_count` (unlike here), so only the base index needs to change.

src/package_io/input_parser.star
```diff
@@ -1779,6 +1782,16 @@ def parse_network_params(plan, input_args):
             + " is not supported, it can only be mainnet or minimal"
         )
 
+    # Base index for builder keys: the first one no genesis validator has claimed.
+    # preregistered_validator_count can exceed the participants' validator sum, and deriving
+    # builders inside that range duplicates a validator's key, which fails genesis. Computed once
+    # here so genesis, the log line and buildoor cannot drift apart. Unconditional: buildoor
```
iurii-ssv, line 1788: Accuracy nit tied to the inline note in `main.star`: "buildoor cannot drift apart" isn't fully true yet — the deprecated `mev_type: buildoor` path at [`main.star:700`](https://github.com/shane-moore/ethereum-package/blob/c3b2cbfadd661d84b3e4bc9597f4c863f4972d17/main.star#L700) still derives from `total_validator_count`. Fixing that line makes the claim hold; otherwise worth softening the wording.

main.star
shane-moore: Good catch, taken. `8907693` swaps that site to `builder_key_start_index` (base only, keeping the path's aliasing of the first genesis builder key) and deletes the now-dead `total_validator_count` accumulator, so no consumer can drift from the genesis registration again.

src/package_io/input_parser.star
shane-moore: Resolved by `8907693`: with the last `total_validator_count` consumer swapped and the variable deleted, the claim holds as written.

iurii-ssv (approved):

pk910 (approved):

barnabasbusa (approved):

PR #1476, opened 2026-08-12
fix: Enable buildoor for Nimbus

ahshum:


src/package_io/input_parser.star
```diff
@@ -2692,8 +2692,10 @@ def apply_buildoor_payload_attributes_flags(participant):
...
             "--Xfork-choice-updated-always-send-payload-attributes=true"
         )
+    elif participant["cl_type"] == "nimbus":
+        # nimbus does not require flags to emit payload_attributes.
+        pass
     else:
-        # nimbus has no flag to emit payload_attributes.
         fail(
             "buildoor requires the CL feeding it to be one of "
             + "[lodestar, prysm, lighthouse, grandine, consensoor, teku]: '{0}' has no flag to build a payload on each slot ".format(
```
pk910: nimbus should be added here too :)

src/package_io/input_parser.star
barnabasbusa: We probably don't even need the else, as now all clients support it.
```suggestion
    elif participant["cl_type"] == "nimbus":
        # nimbus does not require flags to emit payload_attributes.
        pass
```

barnabasbusa (approved):

PR #1480, opened 2026-08-24
feat: derive builder BLS keys from a dedicated builder mnemonic

barnabasbusa:
## Summary
Fixes the buildoor/builder key collision reported today, by removing its root cause instead of patching the offset.

**Dedicated builder mnemonic** — builder BLS keys previously derived from `preregistered_validator_keys_mnemonic`, which required `builder_key_start_index = max(actual_num_validators, preregistered_validator_count)` to dodge validator keys. That base shifted with validator topology and the `mev_type: buildoor` path additionally derived at the *first genesis builder's slot*, so buildoor ran behind the same pubkey as a genesis builder / external sidecar. Now:

- new `builder_keys_mnemonic` network param, defaulting to a new `DEFAULT_BUILDER_MNEMONIC` constant (distinct from the validator mnemonic, so validator collisions are structurally impossible)
- genesis-registered builders occupy indices `0..builder_count-1` on that mnemonic
- every buildoor (mev_type and dedicated instances) gets the next index after them via one shared counter, so buildoors can't collide with genesis builders or each other
- `builder_key_start_index` is removed entirely; the builder BLS key is a pure signing identity so the separate mnemonic needs no funding
- external sidecars get a stable identity: builder mnemonic, index 0 — independent of validator counts

**Spec value check** — consensus-specs master (v1.7.0-alpha.14) added `EIP8321_FORK_VERSION`/`EPOCH`, which egg intentionally does not render into the generated config (same handling as EIP7928); excluded from the upstream comparison so the "Check consensus spec values" jobs pass again.

## Breaking change
Tooling that derived builder keys from the validator mnemonic at the old computed offset must switch to `builder_keys_mnemonic` (printed in the run summary) at index 0+.

## Test plan
- `kurtosis lint` clean
- CI: mev-buildoor / buildoor-instances jobs exercise both allocation paths

### DIFF
--- a/src/package_io/constants.star
+++ b/src/package_io/constants.star
@@ -147,6 +147,9 @@ MEV_BOOST_SERVICE_NAME_PREFIX = "mev-boost"
 COMMIT_BOOST_SERVICE_NAME_PREFIX = "commit-boost"
 MEV_BOOST_PORT = 18550
 DEFAULT_MNEMONIC = "giant issue aisle success illegal bike spike question tent bar rely arctic volcano long crawl hungry vocal artwork sniff fantasy very lucky have athlete"
+# Mnemonic for builder BLS keys, deliberately distinct from DEFAULT_MNEMONIC so
+# builder keys can never collide with validator keys.
+DEFAULT_BUILDER_MNEMONIC = "baby envelope toddler valid pottery buddy cash spare such hedgehog ring ramp item seminar rely select advance knife cruel cereal left father model tissue"
 
 PRIVATE_IP_ADDRESS_PLACEHOLDER = "KURTOSIS_IP_ADDR_PLACEHOLDER"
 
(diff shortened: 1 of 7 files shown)


src/package_io/constants.star
```diff
@@ -147,6 +147,9 @@ MEV_BOOST_SERVICE_NAME_PREFIX = "mev-boost"
 COMMIT_BOOST_SERVICE_NAME_PREFIX = "commit-boost"
 MEV_BOOST_PORT = 18550
 DEFAULT_MNEMONIC = "giant issue aisle success illegal bike spike question tent bar rely arctic volcano long crawl hungry vocal artwork sniff fantasy very lucky have athlete"
+# Mnemonic for builder BLS keys, deliberately distinct from DEFAULT_MNEMONIC so
+# builder keys can never collide with validator keys.
```
barnabasbusa:
```suggestion
```

pk910 (approved):

PR #1485, opened 2026-09-05
feat(cl): add crysm consensus client

potuz:
Adds `crysm`, a C++ consensus client, as a `cl_type`.

## Shape of the change

A launcher plus registrations, following the consensoor precedent:

- `src/cl/crysm/crysm_launcher.star` — the launcher
- `src/package_io/constants.star` — `crysm` in `CL_TYPE` and `VC_TYPE`
- `src/package_io/input_parser.star` — both image tables, and the in-process-VC default list
- `src/vc/vc_launcher.star` — the `return None` hatch
- `src/cl/cl_launcher.star` — import and dispatch entry
- `.github/tests/crysm-basic.yaml` — one participant, Gloas at genesis, dora

## The validator arrangement

crysm's validator client is the same binary as the beacon node, talking to itself over the Beacon
API on loopback. So there is no VC service to launch: `crysm` is in `VC_TYPE` only to reach the
`return None` hatch, exactly as `consensoor` is, and `use_separate_vc` defaults to `False` for it.

The keys are passed as `--keystores-dir` and `--secrets-dir`. An earlier version of this launcher
rendered a `settings.json` instead, which no other CL launcher does; crysm grew the two flags so this
diff looks like its neighbours.

## crysm is Gloas-only

It carries no pre-Gloas containers — `forks::gloas` is the only fork tag in its type system — so it
can start only where Gloas is at genesis. Two consequences worth stating plainly:

- Its coverage is one file, not the matrix the mature clients enjoy. The ~60 pre-fork test files are
  closed to it.
- **It cannot be added to any existing test file.** None of them sets `gloas_fork_epoch: 0`, so
  folding crysm into `mix-with-tools.yaml` or similar would mean changing that file's fork schedule
  for every other client in it. That is why `crysm-basic.yaml` is a new file, and why this PR adds an
  eighth entry to the `per-pr.yml` matrix rather than reusing one of the seven. I appreciate that
  matrix is deliberately short; I could not find a way to exercise a Gloas-only client without it,
  and am happy to drop the entry and take nightly-only coverage if you would rather.

`crysm-basic.yaml` pins `ethpandaops/geth:glamsterdam-devnet-8`, `ethpandaops/dora:master` and
`ethereum-genesis-generator:6.2.0`. Each is load-bearing for a Gloas-at-genesis enclave and each is
commented in the file.

## The image

`ghcr.io/offchainlabs/crysm:stable` and `:stable-minimal`, published by OffchainLabs — the same
arrangement as `offchainlabs/prysm-beacon-chain:stable`, so this needs nothing from your image
pipeline. Both are multi-arch (`linux/amd64` + `linux/arm64`), built on native runners.

## A latent bug this touches but does not fix

`cl_shared.append_bootnode_arg` joins bootnode ENRs without filtering empties:

```python
bootnode_arg = ",".join([ctx.enr for ctx in bootnode_contexts[: constants.MAX_ENR_ENTRIES]])
```

Any client returning `enr=""` therefore injects a blank entry into the bootnode list handed to
lighthouse, teku, lodestar and grandine; `nimbus_launcher.star` and `prysm_launcher.star` have the
same unguarded loop. Because `""` is not `None`, the devnet ENR-list fallback is skipped too, and
nothing errors — the enclave simply never peers while every service reports healthy.

An early version of this launcher had exactly that defect. It now returns a real ENR, so nothing in
tree triggers this today and the PR leaves the shared helper alone. Happy to send a separate one-line
hardening PR if you want it.

## Known limitations

- **Checkpoint sync over HTTPS does not work.** crysm's `--checkpoint` takes `host:port` and does no
  TLS, so the launcher strips the scheme. Fine against an in-enclave checkpointz, not against a
  public provider.
- **No `tcp-discovery` port.** crysm speaks QUIC only, so the launcher builds its own public-port map
  rather than calling `cl_shared.get_general_cl_public_port_specs()`.
- **No `/eth/v2/beacon/blocks/{block_id}` or rewards endpoints yet**, which some tooling reads.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/tests/crysm-basic.yaml
+++ b/.github/tests/crysm-basic.yaml
@@ -1,32 +1,22 @@
-# crysm is Gloas-only -- it carries no pre-Gloas containers -- so it can appear only where Gloas is
-# at genesis. That is why this file cannot be folded into one of the mix suites, none of which fork
-# to Gloas at all.
 participants:
-  # devnet-8 rather than a stock geth: Gloas at genesis needs an execution client that speaks ePBS.
   - el_type: geth
     el_image: ethpandaops/geth:glamsterdam-devnet-8
     cl_type: crysm
-    # One binary, no separate VC service, which is also why crysm has no VC_TYPE launcher.
     use_separate_vc: false
     validator_count: 128
 
 network_params:
   gloas_fork_epoch: 0
   preset: mainnet
-  # geth rejects a BPO after amsterdam, which Gloas at genesis puts at genesis too.
   bpo_1_epoch: 0
   bpo_2_epoch: 0
 
 additional_services:
   - dora
 
-# Not dora's `latest`: that one decodes ExecutionRequests as three lists where Gloas has five, so it
-# indexes no block and shows an empty chain.
 dora_params:
   image: ethpandaops/dora:master
 
-# The generator has no glamsterdam-devnet-8 tag -- the series stops at devnet-7 -- so this is the
-# release that was current when devnet-8 was generated.
 ethereum_genesis_generator_params:
   image: ethpandaops/ethereum-genesis-generator:6.2.0
 
--- a/.github/tests/crysm-basic.yaml
+++ b/.github/tests/crysm-basic.yaml
@@ -1,0 +1,40 @@
+# crysm is Gloas-only -- it carries no pre-Gloas containers -- so it can appear only where Gloas is
+# at genesis. That is why this file cannot be folded into one of the mix suites, none of which fork
+# to Gloas at all.
+participants:
+  # devnet-8 rather than a stock geth: Gloas at genesis needs an execution client that speaks ePBS.
+  - el_type: geth
+    el_image: ethpandaops/geth:glamsterdam-devnet-8
+    cl_type: crysm
+    # One binary, no separate VC service, which is also why crysm has no VC_TYPE launcher.
+    use_separate_vc: false
+    validator_count: 128
+
+network_params:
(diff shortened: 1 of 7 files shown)


barnabasbusa:
can you yeet your excessive comments please?

potuz:
> can you yeet your excessive comments please?

Yes sir, remove all comments, use the same AGENTS.md from the crysm repo

potuz:
Done in 0423f70 — all comments gone from `crysm_launcher.star` and `crysm-basic.yaml`. The diff adds no comment lines at all now.

One correction while doing it: my first pass stripped comments file-wide and took out your commented-out `check_optimism_package` job in `per-pr.yml`, which was none of my business. Restored — that file is back to a single added line, the matrix entry.

On that entry, since it is the one thing here I would push back on myself: crysm is Gloas-only and no file in `.github/tests/` sets `gloas_fork_epoch: 0`, so it cannot be folded into `mix-with-tools.yaml` or any of the other five without changing that file's fork schedule for every client in it. That is why it is a new file and an eighth matrix entry rather than a participant added to an existing one. I know the matrix is deliberately short — say the word and I will drop the entry and take nightly-only coverage instead.

potuz:
> Done in [0423f70](https://github.com/ethpandaops/ethereum-package/commit/0423f70cf9e339c1b2108895639c828734b4bcb3) 


Why can't you simply reply with the above line that was more than enough... it looks as though you charge by the token

PR #1486, opened 2026-09-08
fix(shadowfork): resume snapshot downloads by byte offset

qu0b:
## Problem

The shadow-fork downloader streams each EL snapshot through a single `curl | tar` with no resume, retried three times **from byte zero**, behind a 6h `plan.wait`. That never completes at mainnet scale.

Observed on a mainnet shadow fork (erigon + reth @ 25852100):

```
Downloading snapshot (attempt 3)
/*stdin*\ : Read error (39) : premature end
tar: Unexpected EOF in archive
snapshot download failed (attempt 3)
...
Caused by: Recipe execution timed-out waiting for the recipe execution to become valid on service 'shadowfork-el-1-erigon-lighthouse'. Tried '21600' times.
```

Six hours, nothing on disk. The public snapshot server drops a connection roughly every 50 GB, so even on a clean link the 370 GB erigon and 815 GB reth snapshots cannot arrive in one stream — and restarting from zero makes every drop cost the whole download.

## Fix

Resume by byte offset instead of restarting. A FIFO counts the bytes `tee` handed the extractor, and each retry continues from exactly there with `curl -C <offset>`. Nothing but the extracted datadir ever touches disk (no tarball), which is what lets erigon + reth fit next to each other on one host at all.

- `curl -C <offset>` exits 33 rather than restarting from zero if the server ignores the Range header, so a non-ranging server fails loudly instead of silently corrupting the stream.
- 22 (HTTP error) and 33 are fatal; every other curl exit resumes after 5s, capped at 500 attempts.
- The readiness wait goes 6h → 24h: two mainnet snapshots share one uplink and 6h was not enough. The wait returns as soon as `/tmp/finished` exists, so a longer cap costs nothing on small networks.
- The inline string-concat command is now a readable script constant; rendered output passes `shellcheck -s sh` and `sh -n` under alpine 3.19.

## Verification

1. **Mainnet, live** (same script, applied as a patch on `36f7886`): both downloaders have already resumed twice — erigon at byte 101,693,456,384 and reth at 100,166,205,440 are each on attempt 3 and still streaming, which the old code could not have survived.
2. **End-to-end on this branch**: `hoodi-shadowfork` @ 240000, one geth participant, `network_sync_base_url` pointed at a rate-limited nginx reverse proxy in front of `snapshots.ethpandaops.io`; the proxy was killed mid-stream:
   ```
   snapshot is 2776250259 bytes
   fetching from byte 0 (attempt 1)
   stream broke (curl 18) at byte 1366425600, resuming
   fetching from byte 1366425600 (attempt 2)
   ```
   `kurtosis run` exit 0, `eth_blockNumber` = 0x3a980 (240000).
3. **Byte-identical extraction** with two forced breaks against a locally served tarball: `sha256` over every extracted file matches a straight extraction.
4. `kurtosis lint` clean.
5. **Non-ranging server fails loudly**: same script behind an nginx proxy that strips `Range`, connection killed mid-stream:
   ```
   stream broke (curl 18) at byte 643997297, resuming
   fetching from byte 643997297 (attempt 2)
   fatal curl error 33 at byte 643997297
   tar: Unexpected EOF in archive
   ```
   exit code 2, no `/tmp/finished` — nothing is silently restarted from zero.
6. Full 5-EL `hoodi-shadowfork` (geth, reth, nethermind, besu, erigon) launched through this branch on the same host — see follow-up comment.


qu0b:
Follow-up verification, all on this branch as checked out (`88c3a3b`, no local modifications):

**Full 5-EL `hoodi-shadowfork` @ 240000** (geth, reth, nethermind, besu, erigon; `glamsterdam-devnet-8` images), snapshots from `snapshots.ethpandaops.io` directly:

```
shadowfork-el-1-geth-lighthouse        snapshot is 2776250259 bytes  fetching from byte 0 (attempt 1)
shadowfork-el-2-reth-lighthouse        snapshot is 2375768854 bytes  fetching from byte 0 (attempt 1)
shadowfork-el-3-nethermind-lighthouse  snapshot is 2594188267 bytes  fetching from byte 0 (attempt 1)
shadowfork-el-4-besu-lighthouse        snapshot is 2623205210 bytes  fetching from byte 0 (attempt 1)
shadowfork-el-5-erigon-lighthouse      snapshot is 5407567239 bytes  fetching from byte 0 (attempt 1)
```

`kurtosis run` exit 0 in 3m24s from launch to all five ELs up. Pre-fork sample: every client advanced 25 blocks over 25 slots on the same head. (Post-Gloas the execution chain stops, which is the known EIP-8282 predeploy issue on shadow forks and unrelated to this change.)

**Mainnet erigon + reth @ 25852100**, same script, still streaming on the same host while the above ran:

```
shadowfork-el-1-erigon-lighthouse  stream broke (curl 92) at byte 154221666432 → resumed (attempt 4)
                                   stream broke (curl 92) at byte 200793310272 → resumed (attempt 5)
                                   ... attempt 6 from byte 244497464576
shadowfork-el-2-reth-lighthouse    stream broke (curl 92) at byte 150002008064 → resumed (attempt 4)
                                   stream broke (curl 92) at byte 206554192128 → resumed (attempt 5)
                                   ... attempt 6 from byte 256618855232
```

Every break is `curl 92` (HTTP/2 stream closed by the server), roughly every 50 GB. That is the failure the old three-attempts-from-zero loop could not get past.

pk910 (approved):

qu0b:
Pushed a follow-up: the byte counter is now curl's own `-w '%{stderr}%{size_download}'` instead of a FIFO + `tee` + `wc`. On the live mainnet run the busybox `wc` was at 82% CPU and `tee` at 45% per stream — more than `tar` + `zstd` together — and there is no reason for it, curl already knows what it delivered. Same semantics, re-verified on the rewritten script:

```
fetching from byte 0 (attempt 1)
stream broke (curl 18) at byte 607420416, resuming
fetching from byte 607420416 (attempt 2)
stream broke (curl 18) at byte 1209532416, resuming
fetching from byte 1209532416 (attempt 3)
→ exit 0, sha256 over every extracted file identical to a straight extraction
```
```
stream broke (curl 18) at byte 297762816, resuming
fetching from byte 297762816 (attempt 2)
fatal curl error 33 at byte 297762816     (server ignoring Range → exit 2, no /tmp/finished)
```

PR #1487, opened 2026-09-08
fix(shadowfork): count delivered snapshot bytes with curl -w instead of a FIFO

qu0b:
Follow-up to #1486, which merged before this landed on its branch.

The byte counter in the resumable snapshot download was a FIFO + `tee` + `wc -c`. On a live mainnet shadow fork (erigon + reth, two concurrent streams at ~100 MB/s each) busybox `wc` sat at 82% CPU and `tee` at 45% per stream — more than `tar` + `zstd` together. curl already knows how many bytes it delivered: `-w '%{stderr}%{size_download}'` reports it on stderr, so the stream stays clean and the FIFO, `tee`, `wc` and the background job all go away. Same resume semantics, fewer moving parts.

Re-verified on the rewritten script:

```
fetching from byte 0 (attempt 1)
stream broke (curl 18) at byte 607420416, resuming
fetching from byte 607420416 (attempt 2)
stream broke (curl 18) at byte 1209532416, resuming
fetching from byte 1209532416 (attempt 3)
→ exit 0, sha256 over every extracted file identical to a straight extraction
```
```
stream broke (curl 18) at byte 297762816, resuming
fetching from byte 297762816 (attempt 2)
fatal curl error 33 at byte 297762816      (server ignoring Range → exit 2, no /tmp/finished)
```

`kurtosis lint` clean; rendered script passes `shellcheck -s sh` and `sh -n` under alpine 3.19 (curl 8.14).


qu0b:
Closing data point from the mainnet run that motivated #1486 (erigon + reth @ 25852100, both snapshots pulled from `snapshots.ethpandaops.io` concurrently through one uplink):

| client | tarball | on disk | attempts | server-side stream closes (`curl 92`) | wall clock |
|---|---:|---:|---:|---:|---:|
| erigon | 370 GB | 465 GB | 8 | 7 | 80 min |
| reth | 815 GB | 1,014 GB | 16 | 15 | 2 h 38 min |

22 dropped connections between the two — every one resumed at the exact byte, both datadirs complete, enclave launched and the ELs answering at 25852100. The old three-attempts-from-zero loop had failed the same run after 6 h with nothing on disk.

qu0b:
Added one more commit: `--speed-limit 1000000 --speed-time 60` on the curl. curl only times out the *connect*; a stream that stalls without closing (no FIN/RST) would otherwise hang the download forever. Under 1 MB/s for 60 s curl exits 28 and the resume loop continues from the delivered byte. No effect on a healthy stream.

qu0b:
Correction on the stall guard: 1 MB/s over 60 s would abort a *slow* link, not just a stalled one — on a congested uplink it would churn every minute and give up after 500 attempts where the plain loop would have crawled to completion. Changed to `--speed-limit 1024 --speed-time 120`: zero progress for two minutes, which a working link never trips. Nothing we observed needed the guard (every break so far was a server-side stream close, which curl exits on); it exists for a connection that stays open and sends nothing, which inside a 24h wait would otherwise be invisible for a day.

pk910 (approved):

qu0b:
Good catch on the empty `/tmp/got` path, and thanks for exercising the exit codes. Fixed in 3d9b13da, but deliberately **not** with a default-to-0: if curl ever exits without its write-out, the realistic cause is a signal kill mid-transfer, and by then the bytes it delivered are already inside `tar`. Resuming at the old offset would replay those bytes into the archive and corrupt the extraction silently. So a missing or non-numeric count is now an explicit fatal:

```
curl exited 137 without a byte count ('Killed') -- bytes already handed to tar are unaccounted for, refusing to resume
```

Reproduced with a `curl` shim that writes 4 KiB and `kill -KILL`s itself: the loop aborts with that line, tar fails, exit 2, no `/tmp/finished` — instead of the ash `arithmetic syntax error`. The pre-transfer exits (7/28/35/…) still carry a `0` write-out and resume as before.

PR #1488, opened 2026-09-08
fix(shadowfork): a missing snapshot byte count is fatal, not zero

qu0b:
Follow-up to #1487 — this commit was pushed to that branch ten minutes after it merged, so main still has the path the review flagged.

`off=$((off + $(cat /tmp/got)))` dies with an ash `arithmetic syntax error` if `/tmp/got` is empty. The review suggested defaulting to 0 and resuming; that would be unsafe. The realistic way curl exits without its `-w` write-out is a signal kill mid-transfer, and by then the bytes it delivered are already inside `tar`. Resuming at the old offset would replay them into the archive and corrupt the extraction silently. So a missing or non-numeric count is now an explicit fatal:

```
curl exited 137 without a byte count ('Killed') -- bytes already handed to tar are unaccounted for, refusing to resume
```

Reproduced with a `curl` shim that writes 4 KiB and `kill -KILL`s itself: the loop aborts with that line, tar fails, exit 2, no `/tmp/finished`. Pre-transfer exits (7/28/35/…) still carry a `0` write-out and resume as before. `kurtosis lint` clean; rendered script passes `shellcheck -s sh` and `sh -n` under alpine 3.19.


qu0b:
Second commit on this PR, from production an hour later: the edge answered a resume's `Range` request with a plain 200 once in ~40 resumes, curl exited 33, and the loop's "fatal on 33" threw away a 367 GiB mainnet geth extraction. curl writes no body on 33 (nor on 22), so the delivered offset is intact and retrying the same offset is safe; only the missing-count case stays fatal. A server that never honours Range still ends at the attempt cap instead of silently restarting from zero.

qu0b:
Bump — green (15/15) and unchanged since 2026-09-08, blocked only on `REVIEW_REQUIRED` (the bot approval doesn't satisfy it).

Worth merging because main is currently *worse* than pre-#1486 in one narrow case: `22|33) fatal` aborts the whole extraction when the edge answers a ranged request with a plain 200 — which it does roughly once in 40 resumes, and which cost a 393 GB mainnet geth extraction the day #1487 landed. curl writes no body on 33, so retrying the same offset is safe; only a missing byte count stays fatal.

qu0b:
Added a third commit answering the question from #pr-review about snapshot servers that don't support resume.

Retrying a 200-to-Range is right for the *transient* case — the edge in front of `snapshots.ethpandaops.io` does it roughly once in 40 resumes, and treating it as fatal cost a 393 GB extraction. But a server with **no** Range support does it on every resume, so the retry turned a certain failure into 500 attempts (~42 min of futility). Consecutive 200s are now counted: ten in a row ends the download with

```
server answered 10 ranged requests with 200: no Range support, cannot resume
```

One or two still retry, so the transient case is unaffected. Note such a server works fine until the first dropped connection — the initial request carries no `Range` — so this only changes what happens at the point resume becomes necessary.

Verified with a `Range`-stripping proxy plus one forced break: exit 2 after 12 s and 11 attempts, naming the cause, instead of the attempt cap.

The same loop, with this bound, is now proposed for the other two consumers of these snapshots: ethpandaops/ansible-collection-general#587 (`snapshot_fetcher`, which used `wget --tries=0 -O -`) and ethpandaops/platform#430 (the ethereum-node snapshot init container, same pattern).

qu0b:
Production evidence for this change, from a mainnet geth pull happening right now (1,079.8 GiB compressed, ~3 h in):

```
[14:03:19] stream broke (curl 92) at 970.5 GiB, resuming (attempt 23)
[14:03:24] stream broke (curl 33) at 970.5 GiB, resuming (attempt 24)
[14:09:44] stream broke (curl 92) at 1011.3 GiB, resuming (attempt 25)
```

The edge answered the resume of attempt 24 with a plain 200 — `curl 33`, no body written — and the loop retried the same offset and carried on to 1,011 GiB and beyond. On `main` as it stands, that same event is fatal and discards 970 GiB of extraction after three hours.

One occurrence in 25 resumes, which matches the ~1-in-40 rate I quoted from the run that first hit it. Also note the shape: a 33 immediately following a 92 at the *identical* offset, i.e. the retry of a genuinely broken stream is what draws the bad 200 — so a server with real Range support still produces these, and treating 33 as "no Range support" without counting consecutive occurrences would be wrong. That is why the third commit bounds it at ten in a row rather than one.

qu0b:
That mainnet pull finished — final numbers, and they revise the rate I quoted upwards.

1,079.8 GiB compressed in 3 h 17 min, **26 resumes**, of which:

```
24 x curl 92   HTTP/2 stream closed by the server
 2 x curl 33   ranged request answered with a plain 200
```

So the 200-to-Range is **2 in 26 here, and ~1 in 22 across every mainnet pull I have measured** — not the 1-in-40 I estimated earlier. Both occurrences would have been fatal on `main` as it stands, one of them at 970 GiB after three hours.

Cut spacing is 1 per 45 GB, and it did not improve when a faulty switch on our side was replaced the same day (1 per 51 GB before, ~20% more per-stream throughput after). Consistent with the cuts being the edge's rather than any client's or link's — they reproduce from an unrelated network too.

